PR #16, opened 2015-04-21
Removed old Docker logging and implemented --log-driver=syslog

femueller:
Added: https://docs.docker.com/reference/run/#logging-driver-syslog


runtime/opt/zalando/runtime/Docker
```diff
@@ -55,7 +55,7 @@ def main(args):
     with open(args.config) as fd:
         config = yaml.safe_load(fd)
 
-    cmd = ['docker', 'run', '-d']
+    cmd = ['docker', 'run', '-d', '--log-driver=syslog']
```
femueller: @hjacobs  Can you please review this particular change?

PR #24, opened 2015-05-08
Feature/support for volumes

lmineiro:
This proposal adds support for automatic attachment of EBS volumes and RAID volumes.

I propose a separate "volumes" config key for these while the "mounts" config key is only used to mount partitions. I also suggest a refactor of the collection of "devices" to a single "partition" key.

The test was extended to test, at least, attachment and creation of some RAID volumes


hjacobs:
@lmineiro please update your Pull Request to allow automatic merge

runtime/opt/zalando/init.d/10-prepare-disks.py
```diff
@@ -53,56 +83,134 @@ def is_mounted(mountpoint):
...
+    else:
+        call = ["mdadm",
+                "--build", raid_device,
+                "--level=" + str(raid_config.get("level")),
+                "--raid-devices=" + str(num_devices)]
+        for device in devices:
+            tries = 0
+            # Give devices some time to be available in case they were recently attached
+            while tries < 3 and not os.path.exists(device):
+                logging.error("Waiting for %s to stabilize", device)
```
hjacobs: can you configure "logging" and use the appropriate log level? (I guess INFO would be fine here)

hjacobs:
BTW, thanks for the pull request! :-)

hjacobs:
Can you also check all your Python files with "flake8" please?
(just ignore the default line length of 80, we use a max length of 120)

runtime/opt/zalando/init.d/10-prepare-disks.py
lmineiro: Will do

femueller:
@lmineiro :+1:

lmineiro:
$ flake8 --max-line-length=120 .
$
lgtm. Any special settings?

PR #28, opened 2015-05-19
Clarify which shell is needed to avoid portability issues

elgalu:
I run into some issues while running create-ami.sh because the script doesn't specify which shell is compatible, is this a bash script or zsh or something else?
Are you guys running it on Linux or OSX only?
Thanks


femueller:
What OS and shell are you running? We've been running it on OSX (zsh) and Ubuntu (bash) without any issues.

PR #38, opened 2015-05-28
add support for specifying options when mounting filesystems

alexeyklyukin:
When mounting databases, we want to specify options like nobarrier or noatime to increase FS performance.

### DIFF
--- a/runtime/opt/zalando/init.d/10-prepare-disks.py
+++ b/runtime/opt/zalando/init.d/10-prepare-disks.py
@@ -113,13 +113,15 @@ def format_partition(partition, filesystem="ext4", initialize=False, is_already_
         LOG.info("Nothing to do for disk %s", partition)
 
 
-def mount_partition(partition, mountpoint, dir_exists=None, is_mounted=None):
+def mount_partition(partition, mountpoint, options, dir_exists=None, is_mounted=None):
     """Mounts formatted disks provided by /etc/taupage.yaml"""
-    if is_mounted is False and dir_exists is False:
-        os.makedirs(mountpoint)
-        subprocess.check_call(["mount", partition, mountpoint])
-    elif is_mounted is False and dir_exists is True:
-        subprocess.check_call(["mount", partition, mountpoint])
+    if is_mounted is False:
+        if dir_exists is False:
+            os.makedirs(mountpoint)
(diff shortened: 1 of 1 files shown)


runtime/opt/zalando/init.d/10-prepare-disks.py
```diff
@@ -113,13 +113,15 @@ def format_partition(partition, filesystem="ext4", initialize=False, is_already_
...
     """Mounts formatted disks provided by /etc/taupage.yaml"""
-    if is_mounted is False and dir_exists is False:
-        os.makedirs(mountpoint)
-        subprocess.check_call(["mount", partition, mountpoint])
-    elif is_mounted is False and dir_exists is True:
-        subprocess.check_call(["mount", partition, mountpoint])
+    if is_mounted is False:
+        if dir_exists is False:
+            os.makedirs(mountpoint)
+        if options and len(options) > 0:
```
hjacobs: you can leave "len(options) > 0" away, as options only is true if the list is non-empty

runtime/opt/zalando/init.d/10-prepare-disks.py
```diff
@@ -113,13 +113,15 @@ def format_partition(partition, filesystem="ext4", initialize=False, is_already_
...
+            subprocess.check_call(["mount", "-o", options.replace(' ',''), partition, mountpoint])
```
hjacobs: would be good to stay DRY: only call mount once and prepare/add arguments when necessary

runtime/opt/zalando/init.d/10-prepare-disks.py
alexeyklyukin: Removed, thank you!

runtime/opt/zalando/init.d/10-prepare-disks.py
alexeyklyukin: Makes sense, changed the code to prepare the list of arguments and call mount only once.

hjacobs:
Travis CI build fails: ./runtime/opt/zalando/init.d/10-prepare-disks.py:123:60: E231 missing whitespace after ','

(flake8)

alexeyklyukin:
Doh, added that space :-)

PR #41, opened 2015-05-29
Change permissions to world-writable for mount points.

alexeyklyukin:
Docker images should be able to write to the mounted EBS volumes.


alexeyklyukin:
Not for python3 (see https://www.python.org/dev/peps/pep-3127/):

- octal literals must now be specified with a leading "0o" or "0O" instead of "0";

PR #43, opened 2015-06-01
[scalyr] Make log parser configurable (fixes #42)

### LINKED ISSUE #42: [scalyr] Make log parser configurable in user data
The parser to use for the application.log is hardcoded to 'slf4j'. We have different log formats for our applications and would like to introduce multiple parsers for that.

This could be configurable in the user data just like the scalyr api token.

nabcos:
Introduce 'scalyr_application_log_parser' in the user data to set the parser name. Fixes issue #42.


hjacobs:
@nabcos can you also document this property in the documentation please?
Thanks!

femueller:
:+1:

nabcos:
@hjacobs See zalando-stups/documentation#16, is that sufficient?

PR #47, opened 2015-06-03
pass app-id, app-version and instance-id as env variables to docker

harti2006:


runtime/opt/zalando/runtime/Docker
```diff
@@ -29,6 +30,15 @@ def get_env_options(config: dict):
...
         yield 'ETCD_URL=http://172.17.42.1:2379'
 
+    yield '-e'
+    yield 'APPLICATION_ID={}'.format(config.get('application_id'))
+
+    yield '-e'
+    yield 'APPLICATION_VERSION={}'.format(config.get('application_version'))
+
+    yield '-e'
+    yield 'INSTANCE_ID={}'.format(boto.utils.get_instance_metadata()['instance-id'])
```
hjacobs: I believe "INSTANCE_ID" should not be passed automatically every time (what's the use case? if it's needed you can use the EC2 metadata service to fetch it)
In other words: if we pass the instance-id, why don't we pass all other EC2 metadata? IMHO the EC2 metadata should better be written to the /meta folder

sarnowski: And the taupage.yaml is also available in /meta (or should be, if not done yet)

sarnowski: Since these information are all an API for apps, it would also bind us to AWS which we currently do not.

harti2006: I'm using these information in https://github.com/zalando-stups/friboo/pull/15/ to build unique log file names.

I think it's useful for many applications to know about their environment. "INSTANCE_ID" could also have been called "HOST_ID" or whatever. If we do not pass the information to the docker process, but use the EC2 metadata service from inside the application, we would bind all of our apps to AWS. This would IMHO make the vendor locking even worse.

femueller: We could add this as an optional value if you need it and we don't want to
vendor lock taupage.
Am 04.06.2015 07:54 schrieb "André Hartmann" <notifications@github.com>:

Issue #48, opened 2015-06-03
Add --net=host option for starting the docker container


chutium:
we need such a feature for docker command for our Zookeeper appliance

for example we want to do this by senza create:

sudo docker run -p 2181:2181 -p 2888:2888 -p 3888:3888 -d --net=host efd7a9692031

we need to add this --net=host

Thanks

hjacobs:
@chutium IMHO I think your point is not valid --- Zookeeper uses TCP and the ZK instances can connect to each other with TCP without problems, check with the Cassandra guys how they do it (similar cluster problem: Cassandra nodes need to talk with each other)

chutium:
we tried to create distributed zk cluster with this docker option without STUPS toolbox, works fine.

each zk node try to connect themselves according to the address defined in zoo.cfg, without --net=host, the zoo.cfg in docker container must be set as (for example for server node 1):

server.1=localhost:2888:3888
server.2=private_ip2:2888:3888
...
...

but actually set a server location to localhost is wrong.

if we set the setting of server1 in docker container to something like:
server.1=private_ip1:2888:3888
server.2=private_ip2:2888:3888

then the docker container is not able to connect to itself using private IP of host, because docker container uses a virtual interface on host. (https://docs.docker.com/articles/networking/)

so the option --net=host solve this problem.

sarnowski:
You obviously have to use the Server IP/host and not dockers internal IP.

chutium:
yes, i mean server IP (ec2 instance's private IP)

chutium:
problem i want to point out is that, docker container can not connect to

chutium:
i have an idea, maybe we should check the security group... i will give a feedback tomorrow.

drummerwolli:
we tried it again, the problem is nothing with the network config, nor something else. ZK wants to bind a listening address, but since the docker container is listening behind the docker0 address, ZK canÄt bind this address. therefore we need to access the host system's network interface and bind the address and port directly there. see these logs here:

```2015-06-04 08:11:16,535 [myid:6] - ERROR [/172.31.7.6:3888:QuorumCnxManager$Listener@517] - Exception while listening
java.net.BindException: Cannot assign requested address```

172.31.7.6 is our internal IP form the host machine.

**Comment by chutium** [comment]
so @femueller , please reassign or? setting of security group is not the problem.

**Comment by chutium** [comment]
it seems we need to add something like
```
    if config.get('networktype'):
[...]

PR #55, opened 2015-06-04
Add networking option for starting the docker container with different network settings

chutium:
Add networking option for starting the docker container with different network settings

such as to add an option like
--net=host

ref. #48


chutium:
ah, sorry, it was wrong :)

Issue #40, opened 2015-05-28
New Relic Servers


prayerslayer:
Investigate if we want to support [New Relic Docker integration](https://docs.newrelic.com/docs/servers/new-relic-servers-linux/installation-configuration/enabling-new-relic-servers-docker).

femueller:
I've been contacting already NewRelic regarding some auto deregister functionality for Docker containers and EC2 instances.

femueller:
Containers become automatically deregistered after seven days and can be also erased via the NewRelic API if somebody would like to. It's suggested to keep it like this in order to keep historical data otherwise you can remove them on your own using their API

prayerslayer:
Would be cool to have the 7 days timeframe configurable though, because most of the time I really don't care about servers that stopped reporting > 24h ago :)

femueller:
I've requested it at NewRelic, let's see.. :+1:

femueller:
There are still things to clarify. Therefore it won't be merged for now.

apfeiffer85:
After we had a talk with NewRelic we will just test the Agent again and we also try the Docker integration.

PR #62, opened 2015-06-08
Added newrelic-sysmond

femueller:


femueller:
This should be kept open until #40 is closed.

PR #70, opened 2015-06-12
Allow to run docker alongside docker via -v docker.sock

elgalu:
Although this is equivalent to giving the container sudo access to the host it allows to be able to run docker alongside within the container like in a Jenkins environment for example.

It is also an alternative to using `--privileged` for running docker inside docker.

I'll update the docs if you agree with the change.

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -119,6 +119,14 @@ def get_volume_options(config: dict):
         # /opt/taupage/init.d/10-prepare-disks.py will mount the path below "/mounts" on the host system
         yield '{}:{}'.format('/mounts{}'.format(path), path)
 
+    # this is equivalent to giving the container sudo access to the host
+    # and is used to be able to run docker alongside within the container
+    if get_or(config, 'allow_docker_alongside_docker', False):
+        yield '-v'
+        yield '/usr/bin/docker:/usr/bin/docker'
+        yield '-v'
+        yield '/var/run/docker.sock:/var/run/docker.sock'
+
     # meta directory, e.g. containing application credentials retrieved by berry
     yield '-v'
     yield '/meta:/meta'


hjacobs:
Hmm, I'm not sure if running Docker "alongside" Docker is a good idea --- it might conflict with our tooling (assuming there is exactly one Docker container running). Running Docker inside Docker (--privileged) is a cleaner solution as far as I can see right now.

Let me discuss this topic in the team first.

hjacobs:
@femueller @apfeiffer85 could you have a look at this PR and discuss the implications?

femueller:
I would suggest too, that you'll stick to run Docker inside Docker. That's why we've actually allowed to run --privileged

femueller:
@elgalu Would this be a possible solution for you?

elgalu:
With the alongside solution don't need to install stuff in my Dockerfile while with [dind](https://github.com/jpetazzo/dind) I have to. If the implications of this changes are too undesirable will stick with dind then ;)

One bad thing about dind is that senza will easily loose track of what's running in there while with alongside you can later on, for example if you install New Relic docker support. Still in alpha/beta we may evaluate the impact of both options and discard them or add a disclaimer like *use at your own risk as this is not recommended nor supported* then time will tell. Note I want this for testing instances with selenium and Jenkins so if things go bad is not critical while for critical apps they should probably avoid docker in docker or docker alongside docker I suppose.

femueller:
So we'll have to stick with `--privileged` for now as we don't know how we'll move further in the close future.
For now I'll close it.

mfellner:
I have recently run into this problem while trying to set up a CI service that uses Docker containers for test environments. Any continuous delivery service, since it needs to build Docker images, also has the same issue.

Running Docker inside Docker comes with an unnecessarily big overhead and images become bigger than they have to be (e.g. Drone.io which is written in Go can run from a minimal Docker base image but would otherwise require Debian/Ubuntu and the docker.io package). Furthermore, it would be advantageous if the service that depends on Docker can use the host's Docker service in order to cache image layers more effectively.

Taupage is already very restrictive about how containers can be run, while on the other hand some features, like Scalyr logging, are baked in. If there are no major security concerns, I would really like to see this feature in Taupage soon.

sarnowski:
Same discussion for Kubernetes: https://github.com/kubernetes/kubernetes/issues/391

There is this need for CI systems to be able to create Docker containers obviously. Unfortunately, with Docker, this is not that easy possible as you need access to a working Docker daemon (unlike for example with rkt which builds "offline"). The only way is to either give the container access to its own daemon or give the container full root permissions to spawn an own daemon. Both are a security nightmare.

Kubernetes solved that by allowing servers to be started with a special properties, that allows containers to be run in privileged mode. Its a core setting of the server, with the intention, that you have separate servers for your CI and your production system. Production systems should never allow privileged containers.

For Taupage, this could mean either you get access to mint credentials xor you can run privileged containers. Having both would have serious security impacts (what is happening on the server is not auditable anymore).

sarnowski:
As an additional hint for CI configurer: If you are running your worker in privileged mode/access to the daemon and execute user provided code (e.g. running "mvn test"), you have to consider your worker to be compromised. You have to spawn a new server for the next build. If you do not, any job could replace your trusted worker image and run future jobs with a malicous, code injecting service.

So in summary: your server must not have credentials to access the production systems. Every build needs a newly set-up server to not compromise future builds,

PR #71, opened 2015-06-12
Allow to set 'GatewayPorts yes' to reverse tunnel possible

elgalu:
This allows to set reverse tunnels which is useful for selenium testing instances that then can be used by a user to test his local web server for example.

I'll update the docs if you agree with the change.


hjacobs:
I believe the config property should have "ssh_" as a prefix, to make clear it's only for the sshd config.

@elgalu could you explain your use case in a bit more detail please?

elgalu:
I agree with the property name and updated the PR.

My use case is to run selenium in AWS through senza and being able to test my local running app in the cloud, for that I need to open an `ssh -R` reverse tunnel into the AWS taupage instance so the selenium browser running in there can hit my web server at port 3000 for example. Probably an alternative solution would be to setup a VPN inside the instance but that's overkill, reverse tunnel is enough I think.

This also allows to tests an integration server running on my local network at work but use AWS running selenium instances, of course I could do it outside of taupage or just fix the sshd_config after the instance is created but that is one more step that would be nice to avoid upfront.

hjacobs:
ok, I see

hjacobs:
@femueller @apfeiffer85 could you have a look at this PR and discuss the implications?

femueller:
@elgalu :+1: If it enables you. Would you be so kind to update the documentation according to it? I would merge it then.

PR #90, opened 2015-07-08
Fixed typo

femueller:


runtime/opt/taupage/bin/register-in-etcd.py
```diff
@@ -35,7 +35,7 @@ def get_health_check_url(config: dict):
 
 
 def is_healthy(url):
-    '''Return true iff GET on the URL returns status 200'''
+    '''Return true if GET on the URL returns status 200'''
```
hjacobs: this was not a typo! "iff" is the common abbreviation for "if and only if" --- it's used in math to make logical implications clear: https://en.wikipedia.org/wiki/If_and_only_if

hjacobs:
basically using "iff" is shorter than saying "returns true for HTTP status 200 and returns false for all other status codes"

PR #91, opened 2015-07-09
Add Cloudwatch Logs support

marco-hoyer:
We would like to use Cloudwatch Logs as log storage with taupage. The daemon get's installed during build time using the official install script. Autostart is disabled. For the daemon to start, one needs to configure the log files and their corresponding log groups in taupage.yaml. So the behaviour remains the same. There is also a basic integration test verifying that the agent is running and contains the correct configuration.


hjacobs:
good idea! This makes it much easier for people without a separate logging provider account :-)

femueller:
@marco-hoyer Thanks so much for this! Could you please add some information to the [documentation](https://github.com/zalando-stups/documentation) too?
Once this is done, I think we can definitely merge this.

marco-hoyer:
@femueller You have a PR on documentation for this.

tests/spec/base-packages_spec.rb
```diff
@@ -5,6 +5,7 @@
                 'python3-requests',
                 'python3-yaml',
                 'python3-pip',
+                'python3-jinja2',
```
marco-hoyer: @femueller done!

femueller:
Thanks! But it seems like it's not flake8 compatible yet. When this is fixed we can merge it.

marco-hoyer:
Hmm, the failure seems to be in Docker.py, not modified by me...

`./runtime/opt/taupage/runtime/Docker.py:108:83: W291 trailing whitespace`

marco-hoyer:
I see some strange conflicts everytime I try to update the is24 fork against origins master. Will close this and start over cleanly.

Issue #86, opened 2015-06-26
Upgrade to Docker 1.7


hjacobs:
We fixed https://github.com/zalando-stups/taupage/issues/84, so we can upgrade to Docker 1.7 after "some" time (old images in Pier One cannot be used by Docker 1.7)

hjacobs:
This issue is already two weeks old, I think we could now upgrade Taupage to Docker 1.7.

PR #92, opened 2015-07-10
Install latest Docker release with taupage #86

femueller:


build/setup.d/20-install-packages.sh
```diff
@@ -9,7 +9,7 @@ libyaml-0-2
 linux-image-extra-$(uname -r)
 logentries
 logentries-daemon
-lxc-docker-1.6.2
+lxc-docker
```
hjacobs: maybe we should pin the Docker version to 1.7 --- the next Docker version might be broken (for us) again...

femueller: All right, it was just a proposal. Forgot to comment to be honest.

femueller:
Fixes https://github.com/zalando-stups/taupage/issues/86

PR #93, opened 2015-07-10
Add prometheus node exporter

femueller:


hjacobs:
On which port is the node exporter running (important for setting the right security group rules)? How to query the exported metrics?

femueller:
@hjacobs It's running on port 9100 and metrics can be queried like this:
````
curl http://your_server_ip:9100/metrics
````

femueller:
Approach will be switched to rather run the exporter in a Docker container.

PR #94, opened 2015-07-13
Add cloudwatch logs support

marco-hoyer:
We would like to use Cloudwatch Logs as log storage with taupage. The daemon get's installed during build time using the official install script. Autostart is disabled. For the daemon to start, one needs to configure the log files and their corresponding log groups in taupage.yaml. So the behaviour remains the same. There is also a basic integration test verifying that the agent is running and contains the correct configuration.

This is a new and clean PR.


femueller:
:+1:  Thanks for also fixing the flake8 error Marco, which wasn't created by you.

apfeiffer85:
was my fault ;-) But anyway, thanks for the Cloudwatch Logs support.

marco-hoyer:
Thanks for merging! Some other stuff will follow soon ;-)

femueller:
Good stuff @marco-hoyer :) Looking forward to it!

PR #95, opened 2015-07-14
Finally terminate instances on failed builds

marco-hoyer:
If anything fails during build or test of the taupage ami, the used ec2 instances remain active. I propose to cleanup finally if anything went wrong. The traps should do the job.


femueller:
Thanks a lot marco! :+1:

PR #98, opened 2015-07-16
Dynamic docker config

marco-hoyer:
To be able to use private Docker registries, we need basic auth support for docker. The best approach seems to be to make the dockercfg configurable by taupage.yaml. This allows to configure whatever one wants. There is also a test verifying it to work properly.

This PR also includes a fix for the new relic agent install. Due to the nature of how the setup.d scripts are being executed (by sourcing them), an exit would lead to the install process to stop right there if no new relic agent is present.

### DIFF
--- a/build/setup.d/86-add-newrelic-java-agent.sh
+++ b/build/setup.d/86-add-newrelic-java-agent.sh
@@ -7,6 +7,7 @@ newrelic_yaml=/opt/proprietary/newrelic/newrelic.yml
 #check if there is a no_newrelic file and exit
 if [ ! -d /opt/proprietary/newrelic ]; then
 	echo "INFO: no NewRelic agent found, skip configuration"
+else
 
 	# standard config files
 	newrelic_agents="
--- a/build/setup.d/86-add-newrelic-java-agent.sh
+++ b/build/setup.d/86-add-newrelic-java-agent.sh
@@ -7,7 +7,6 @@ newrelic_yaml=/opt/proprietary/newrelic/newrelic.yml
 #check if there is a no_newrelic file and exit
 if [ ! -d /opt/proprietary/newrelic ]; then
 	echo "INFO: no NewRelic agent found, skip configuration"
-else
 
 	# standard config files
 	newrelic_agents="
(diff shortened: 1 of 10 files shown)


apfeiffer85:
we already modify the newrelic-agent register and install process, so if there is no agent present or there is no newrelic_account_key in the .yaml file the script's will stop. 
And the "exit" status is correct because our main setup.sh script stops only on "exit/return 1" status. 
So the Setup Process should continue with an "normal" exit. 

Did you get some Errors if there is no newrelic_account_key in the YAML file ?

marco-hoyer:
We'd seen problems with an older version of the newrelic install script (https://github.com/ImmobilienScout24/taupage/blob/deb68f8618373991536c8d1c60eb96e4a6aebc16/build/setup.d/86-add-newrelic-java-agent.sh). It's exit terminated the install script execution so that no other script was executed after it. Did you test it to work properly?

apfeiffer85:
Yeah, you're right i mixed up the setup.sh with the init.sh script .... so we have to modify the scripts.

apfeiffer85:
We also changed the directory of the Agents so the "no_newrelic" file isn't needed anymore. 
I also changed the Agent scripts (Newrelic and appdynamics), so could you please adjust or PR with the new versions of the scripts. 

Sorry for that ... ;-)

apfeiffer85:
And could you also add this dockercfg stuff to our https://github.com/zalando-stups/documentation ?

marco-hoyer:
I will remove the modification from PR. If you're willing to merge, I will also add some documentation. Wanted to wait until we clarified the suitability of this change.

marco-hoyer:
Ok, our modification at newrelic install script has gone. Will add some documentation now.

marco-hoyer:
doc is done: https://github.com/zalando-stups/documentation/pull/35

femueller:
@apfeiffer85 could you review this part please again?

apfeiffer85:
From my site it's ok to merge now. :-) :+1:

femueller:
@hjacobs what would you say?

hjacobs:
I think it's fine :+1:

femueller:
:+1:  merged

PR #104, opened 2015-07-27
Store taupage yaml in meta

femueller:
@hjacobs Could you please have a look at this, since it's a pretty critical change?

### DIFF
--- a/build/files
+++ b/build/files
@@ -16,7 +16,7 @@
 
 /etc/ssh-access-granting-service.yaml root:root 0444
 
-/etc/taupage.yaml root:root 0444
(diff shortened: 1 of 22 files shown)


apfeiffer85:
i guess it looks good so far.

hjacobs:
At least berry also uses /etc/taupage.yaml path, so right now it would fail.
Maybe just do "double-write" (have both /etc/taupage.yaml and /meta/taupage.yaml) for some time until all dependencies are updated, too.

apfeiffer85:
are there some other tools who uses the /etc/taupage.yaml, if it is only berry we can adjust this, too. :-)

PR #105, opened 2015-07-28
Use --create instead of --build mdadm option when creating a new RAID array. 

alexeyklyukin:
Build option should be used only for compatibility with old systems and does not support all levels, in particular, level 10.


femueller:
:+1: Thanks @alexeyklyukin

PR #111, opened 2015-08-10
Retry 3 times when pushing taupage yaml

prayerslayer:
Otherwise it could be the case that we cannot obtain a valid token due to rate limiting.


femueller:
Thanks :+1:

Issue #114, opened 2015-08-10
Only run awslogs-nanny.sh when needed


hjacobs:
There are frequent syslog messages about awslogs-nanny.sh even if no Cloud Watch logs are used --- awslogs-nanny.sh should run via CRON if Cloud Watch logs are enabled:

    CRON[1681]: (root) CMD (/var/awslogs/bin/awslogs-nanny.sh > /dev/null 2>&1)

femueller:
This was fixed by #115

PR #115, opened 2015-08-10
#114 rename cron.d file to only run awslogs nanny when needed

hjacobs:
Untested, please do not merge yet!

Comments appreciated.


femueller:
Looks good from the idea. I will test and merge it, if everything works.

Issue #116, opened 2015-08-11
Use Token Service only as fallback for non-confidental OAuth clients (client_id = null)


hjacobs:
To avoid one more layer of potential problems (e.g. rate limits) we should circumvent the Token Service in the default case and use the OAuth 2 provider endpoint directly.

Changes are needed in the "get_token" method.

hjacobs:
We can use the existing "tokens" library for this: https://github.com/zalando-stups/python-tokens

prayerslayer:
Fixed with #117

PR #117, opened 2015-08-11
Issue 116: Use OAuth access token URL instead of Token Service by default

hjacobs:
Untested fix for #116, for review

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -7,9 +7,7 @@ import argparse
 import base64
 import boto.kms
 import boto.utils
-import json
 import logging
-import os
 import pierone.api
 import pwd
 import requests
@@ -18,7 +16,7 @@ import subprocess
 import time
 import yaml
 
-from taupage import is_sensitive_key, CREDENTIALS_DIR, get_or, get_default_port
+from taupage import is_sensitive_key, CREDENTIALS_DIR, get_or, get_default_port, get_token
 
 AWS_KMS_PREFIX = 'aws:kms:'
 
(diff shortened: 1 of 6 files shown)


runtime/opt/taupage/runtime/Docker.py
```diff
@@ -186,29 +184,14 @@ def registry_login(config: dict, registry: str):
...
-        credentials = json.load(fd)
-
-    user = credentials.get('application_username')
-    passwd = credentials.get('application_password')
+    token = get_token(config, 'pierone', ['uid'])
 
-    if not user or not passwd:
-        logging.warning('Invalid OAuth credentials: application user and/or password missing in %s', path)
+    if not token or 'access_token' not in token:
+        logging.warning('Missing OAuth token for Pier One login')
```
prayerslayer: What’s the point of warning here? Shouldn’t this rather be an exception as we can’t continue anyways when we want to run a Docker container?

runtime/usr/local/lib/python3.4/dist-packages/taupage/__init__.py
```diff
@@ -126,10 +127,17 @@ def get_config(filename=TAUPAGE_CONFIG_PATH):
...
     token_url = config.get('token_service_url')
 
+    if not oauth_access_token_url:
+        logging.warning('No OAuth access token URL configured in Taupage YAML ("oauth_access_token_url" property)')
+
     if not token_url:
         logging.warning('No token service URL configured in Taupage YAML ("token_service_url" property)')
+
+    if not oauth_access_token_url and not token_url:
+        # neither of the URLs is given, no chance to continue
```
prayerslayer: `raise Exception('Neither token_service_url nor oauth_access_token_url configured')`?

prayerslayer:
Not an expert in Taupage, but looks good to me aside from that I’d fail earlier in some places :+1:

runtime/usr/local/lib/python3.4/dist-packages/taupage/__init__.py
hjacobs: @prayerslayer I think the interface of get_token (as is right now) should return None (return without anything will return None) in case prerequisites might be missing

runtime/opt/taupage/runtime/Docker.py
hjacobs: user's (in a different organization) might still use a Pier One configured without OAuth, so better continue...

prayerslayer: Ah, true, forgot about that option.

runtime/usr/local/lib/python3.4/dist-packages/taupage/__init__.py
```diff
@@ -138,43 +146,61 @@ def get_token(config: dict, token_name: str, scopes: list):
...
+            client_credentials = json.load(fd)
+    except:
+        logging.warning('Invalid OAuth client credentials: could not read %s', client_path)
+        # we might continue as Token Service does not require client credentials
+        client_credentials = {}
+
+    client_id = client_credentials.get('client_id')
+    client_secret = client_credentials.get('client_secret')
+
+    if client_id and client_secret and oauth_access_token_url:
```
prayerslayer: We do not need `client_secret` if the client is non-confidential.

hjacobs:
OK, I will update the PR

Issue #110, opened 2015-08-06
Mount /var/log into the docker container  


apfeiffer85:
make this configurable: 
https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/runtime/Docker.py#L109

Could be a option in the taupage.yaml

hjacobs:
Yes, please make this configurable --- always mounting /var/log may break Docker images relying on internal /var/log folder (e.g. the "Odd Bastion host" image)

PR #123, opened 2015-08-18
mount log directory as read-only

femueller:
This would be a proposal, which should solve the issue we've had. This lets users optionally use ``mount_var_log`` if it is set to ``true``.
@apfeiffer85 or/and @hjacobs please review.

This should fix https://github.com/zalando-stups/taupage/issues/110


femueller:
docs are added here:
http://docs.stups.io/en/latest/components/taupage.html#mount-var-log

PR #124, opened 2015-08-24
This removes the AMI creator ssh key on boot and the ubuntu user. Fix…

femueller:
…es #4


hjacobs:
I would prefer to do it Python style "try .. except ..", i.e. attempt to delete the user and file every time and ignore "file does not exist" errors (e.g. "rm -f") --- IMHO it would be less code and more robust.

femueller:
I will keep it for now like this in order to not run in exit codes, which we don't want to have when an instance starts.

PR #137, opened 2015-10-09
daily rotation of application.log to prevent disk overflows

marco-hoyer:
We've seen disk overflows on instances existing for multiple days with significant traffic leading to significant amount of logs. To prevent this, we would like to rotate application.log on a daily base.


marco-hoyer:
(hopefully) obsoleted by @apfeiffer85

PR #149, opened 2015-10-26
Added mount_certs option

lmineiro:
This adds a new boolean options to taupage config `mount_certs: <true|false>`

This options behaves like the existing `mount_varlog` option.

Since Taupage already bundles a standard SSL certificates tree it helps application maintainers that want to use upstream images like busybox, which are very lightweight, but still need a valid certificate tree to operate correctly.

Most people just take the easy route which is to start from the zalando-ubuntu image but this adds hundreds of unnecessary MBs when the same can be achieved with only a couple MBs. This also helps Pierone

We've tested this approach with our Golang binary, using `FROM busybox:ubuntu-14.04`, which resulted in a 16MB container

I've discussed this with @sarnowski already and I'd like you guys to consider this suggestion.

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -109,10 +109,16 @@ def get_volume_options(config: dict):
     yield '-v'
     # mount the meta directory as read-only filesystem
     yield '/meta:/meta:ro'
+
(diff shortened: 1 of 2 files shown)


femueller:
Sounds like a valid request to me :+1:

PR #150, opened 2015-10-28
create tmpfs for ssh granting service backup filesystem

femueller:


femueller:
This can be merged when the corresponding part in even was done.

PR #152, opened 2015-10-30
added configuration option to allow docker daemon access

sarnowski:


hjacobs:
@femueller @sarnowski please also add this property to the documentation! :scroll:

sarnowski:
That's a good point. The API between client and server is much more fragile and not compatible leading to errors like `client is newer than server (client API version: 1.21, server API version: 1.20)`.

PR #158, opened 2015-11-18
Decrypt newrelic account key if necessary

### LINKED ISSUE #157: newrelic_account_key doesn't support aws:kms encrypted values
[84-register-newrelic-java-agent.sh](https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/84-register-newrelic-java-agent.sh) is missing a check for a aws:kms encrypted value, like [01-register-scalyr-agent.sh#L25](https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/01-register-scalyr-agent.sh#L25):

```bash
# If KMS decrypted, decrypt KMS and save to ACCOUNTKEY variable
if [[ $ACCOUNTKEY == "aws:kms:"* ]]; then
  ACCOUNTKEY=${ACCOUNTKEY##aws:kms:}
  ACCOUNTKEY=`python3 /opt/taupage/bin/decrypt-kms.py $ACCOUNTKEY`
fi
```

Furthermore, `if [[` needs bash, so the shebang should be `#!/bin/bash` instead of `#!/bin/sh`.

mfellner:
* Added kms decryption from [01-register-scalyr-agent.sh](https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/01-register-scalyr-agent.sh#L25)
* Changed shebang to bash

### DIFF
--- a/runtime/opt/taupage/init.d/82-register-newrelic-symond.sh
+++ b/runtime/opt/taupage/init.d/82-register-newrelic-symond.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # read /etc/taupage.yaml
 # get NewRelic Key and register the agent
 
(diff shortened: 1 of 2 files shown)


hjacobs:
@mfellner thanks for the PR :sunny: , I actually thought we would already support this...

@femueller can you check please?

kolja:
@mfellner wants to know: maybe there should be a general way to decrypt any "aws:kms" in the taupage.yaml?

hjacobs:
@kolja @mfellner yes, maybe the parse-yaml.py script could support an (optional) "--decrypt" option to transparently decrypt string values starting with the common "aws:kms:" prefix

sarnowski:
I would also prefer a generic solution that every value, discovered in the initial YAML will be decrypted - independent of the location in the YAML. Then we do not need any special logic for any feature.

whiskeysierra:
> I would also prefer a generic solution that every value, discovered in the initial YAML will be decrypted - independent of the location in the YAML. Then we do not need any special logic for any feature.

:+1:

Issue #143, opened 2015-10-19
Add more Logs for send out to log-providers


drummerwolli:
It would be nice to configure in the Taupage Config Yaml additional Logs in the Docker container, which will be send out seperately to the log providers. Adding other logs with "tail -f" puts them all without overview in one log, which makes it cumbersome to distinguish them (think 10 diferent logs of several applications).

first idea: mount from taupage the location of the log in the docker and tell the log-agent to send those, too.

hjacobs:
I'm personally not a fan of adding this feature to Taupage. One application (Docker image) should usually run only one process and just print to STDOUT. Line prefixes could provide some means of splitting logs.

drummerwolli:
for leightweight applications and rest services, this might apply and i would agree for those. but more heavyweight software might work differently; additionally you as a deployer may have not coded the application (e.g. other OS software, or even proprietary stuff) and you can't adjust the output, etc. E.g. kafka or zookeeper/exhibitor already write 3-4 different logs, sometimes depending on the role the broker/node is taking over. i think databases would also fit in this category ...

this is not a high prio feature request, though. and i personally would like to keep things as simple as possible. but i think having a decent, cleaned up overview about complex systems is more what we aim here for and should be favoured.

marek-obuchowicz:
Lounge team was also considering it, as Elasticsearch logging uses several different files. We're also using JVM GC logging, which, if logged to external file, could be easily tagged/filtered in Scalyr. At the moment it comes out to stdlog and the SLF4j parser in Scalyr fails to correctly handle GC messages (we can't alter logformat of those)

Raffo:
We currently support AppDynamics and it's possible to customize this via: https://github.com/zalando-stups/taupage/blob/ba18617ae8f9a01032d92acaf615f4a39d7f9077/runtime/opt/taupage/runtime/Docker.py#L160-L163

PR #159, opened 2015-11-25
enhanced log locations

drummerwolli:
fixes https://github.com/zalando-stups/taupage/issues/143

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,17 @@
 
 ## Prerequisites
 
-You need to have [jq](http://stedolan.github.io/jq/) and [AWS CLI](http://aws.amazon.com/cli/) preinstalled.
+- [jq](http://stedolan.github.io/jq/) installed
+- [AWS CLI](http://aws.amazon.com/cli/) installed
+- config-stups.sh created/modified (see config-stups-example.sh in the code base and modify it to your needs)
+- adapt other config files
+    - e.g. probably all files in the secret/ folder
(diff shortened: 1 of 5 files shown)


hjacobs:
@apfeiffer85 @femueller let's discuss this PR first before merging (you are the Taupage maintainers)

femueller:
:+1: for a discussion too. We will get this done very soon.

chutium:
any update on this? we really need this, not only for kafka, but also for spark appliance user or Elasticsearch appliance user like @marek-obuchowicz mentioned https://github.com/zalando-stups/taupage/issues/143#issuecomment-149206785, I believe for Spilo or cassandra, pushing multiple log files to scalyr or something is also meaningful.

tail -f is an option, but for a long running service, if the log file rotated (old log file switch to xxx.log.1 or something, and a new xxx.log created), its inode changed, tail -f will not working after that

marek-obuchowicz:
@chutium maybe you can use this library to quickly create a "multiple logfiles tail" tool: http://code.activestate.com/recipes/577968-log-watcher-tail-f-log/ - it handles correctly logfile rotation

apfeiffer85:
We don't want to do this pull request because thats brings a lot of more maintenance to the Taupage system. 
Also the logrotate is not included and therefore it is possible to write the disk full of logs. 

But at the end we need a solution for your case, maybe we can tag the logs and set a different log parser but we have to discuss that with Scalyr or maybe other Logging Providers.

Issue #6, opened 2015-04-01
tune sysctl settings


sarnowski:

femueller:
This could be used as a guide:
http://www.brendangregg.com/blog/2015-03-03/performance-tuning-linux-instances-on-ec2.html

apfeiffer85:
This is also a nice page: http://wiki.mikejung.biz/Sysctl_tweaks

femueller:
I'll close this, because there hasn't been the need so far and we've already added necessary sysctl settings. Further will be added piece by piece if necessary.

PR #160, opened 2015-11-26
Set TCP timeouts explicitly. Fixes #153

### LINKED ISSUE #153: Set TCP timeouts explicitly
The TCP timeout is set to the default: 2h.

```
$ cat /proc/sys/net/ipv4/tcp_keepalive_time 
7200
```

A value in the range of minutes (for example: 1 minute) is desired.

femueller:


femueller:
This also tackles part of https://github.com/zalando-stups/taupage/issues/6

Issue #135, opened 2015-09-30
CloudWatch: monitor other disk partitions


marek-obuchowicz:
Please add the CloudWatch metrics about the disk usage also for EBS mountpoints listed in taupage user-data, if they are provided

femueller:
As a note for this feature: We could basically add a wrapper on top of the provided cloudwatch metrics script, which parses all partitions and adds them automatically to the cronjob.

It could be like this:

    for i in `df -h | grep ^/dev | awk  '{print $6}'`; do echo $i; done

and add it to:

    /etc/cron.d/mon-put-instance-data-cloudwatch (--diskpath=$i)

femueller:
This would be my proposal for "fixing" this issue. As there will be soon a proper release by AWS, which covers this topic I am not really sure if we should integrate this.

femueller:
@marek-obuchowicz What's your opinion about this feature? Do you need this for now? Otherwise I would wait for AWS to integrate it.

femueller:
I'll close this, since nothing has happened here and we're expecting to get this fixed by AWS in the close future.

PR #162, opened 2015-12-03
detect partitions with cloudwatch metrics

femueller:


femueller:
I've closed this, since the issue #135 is also closed.

PR #168, opened 2015-12-29
chmod 600 to mon-put-instance-data-cloudwatch in cron.d

ramirantala:
Currently permissions for the cron.d file are following:

-rw-rw-r-- 1 root root 198 Dec 29 12:46 /etc/cron.d/mon-put-instance-data-cloudwatch

That causes cron execution to fail - actually it is not executed at all.

Dec 29 09:03:02 ip-172-31-170-215 cron[1149]: (*system*mon-put-instance-data-cloudwatch) INSECURE MODE (group/other writable) (/etc/cron.d/mon-put-instance-data-cloudwatch)

Fixed by adding chmod 600


hjacobs:
:+1: LGTM. @femueller or @apfeiffer85 can review & merge.

femueller:
Looks good to me too. I'll just merge it.

Issue #164, opened 2015-12-15
syslog has 8k message limit


messis-rocket:
rsyslog cuts loglines to roughly at 8k, but some application have much more data to show than that (java stack traces etc). 

with $MaxMessageSize set in rsyslog.conf as first definition, the max size can be changed.

can this be added to taupage config?

lixef:
@messis-rocket According to the docs it's default is only [2k](http://www.rsyslog.com/doc/v8-stable/configuration/global/index.html).
Do you have any proposal for a length? Otherwise it would make more sense to me to provide some way to make this option configurable in order to always suit the developers needs.

femueller:
@messis-rocket  According to the docs it's default is only [2k](http://www.rsyslog.com/doc/v8-stable/configuration/global/index.html). Do you have any proposal for a length? Otherwise it would make more sense to me to provide a possibility to make it configurable in order to suit the developers needs.

messis-rocket:
Well the version used in taupage seems to be 7.4.4 and it's manually actually says 4K default: http://www.rsyslog.com/doc/v7-stable/rainerscript/global.html?highlight=maxmessagesize
but I've seen in some other documents that it should be 8K default and that's what I'm seeing the application log. longert lines are pretty much 8111 bytes long. 
I'd like that the max size would configurable, because not all application need to have bigger message size.

femueller:
This feature has been implemented and will be enabled in the next release.

messis-rocket:
Hi, Thank you for this, but looking at the config, it might not work properly. From the document:

Important: In order for this directive to work correctly, it must be placed right at the top of rsyslog.conf (before any input is defined).

femueller:
This is fixed in https://github.com/zalando-stups/taupage/commit/ad035993e3e5fb29f25550dab1938be251e94a84
Thanks for reviewing it :+1:

PR #169, opened 2016-01-07
Enabling to configure MaxMessageSize for ryslog

femueller:
Fixes https://github.com/zalando-stups/taupage/issues/164


messis-rocket:
From document:
Important: In order for this directive to work correctly, it must be placed right at the top of rsyslog.conf (before any input is defined).

Issue #170, opened 2016-01-19
create common audit infrastructure


vetinari:
Currently every team has to create their own way of saving audit log messages from within the container to some "safe storage". This should be built in the taupage image. One way could be to mount  a directory from the host with a socket where the app in the docker container can write to. The messages are read by some script in the host into e.g. /var/log/application-audit.log and the script which sends the /var/log/audit.log content also copies these messages to a central place.

hjacobs:
I'm against integrating that feature into Taupage, Taupage also has enough features --- application specific auditing logs should be handled by the application. Some applications e.g. have databases where audit logs can be stored...

PR #171, opened 2016-01-19
mount /var/log/audit from the host rw to container and push all logs …

vetinari:
…except for those starting with audit from that dir
see https://github.com/zalando-stups/taupage/issues/170


hjacobs:
I'm note sure we should support that, what's the use case? Applications can track their audit logs themselves (most have a database anyway).

femueller:
I agree with Henning here. Theoretically we could add it, but it would also add again a config option for the taupage yaml section.

PR #174, opened 2016-01-28
Node exporter - enable tcpstat collector, add test

marek-obuchowicz:
Adds tcpstat to the default prometheus plugin list, in order to see summary number of TCP connections.
Adds service running requirement to rspec test (it wasn't verified if works fine - test procedure is not described in `README.md`, i wasn't able to run serverspec on my own).


femueller:
Looks good to me :+1: 
Thanks!

PR #182, opened 2016-02-26
Enable bind mounts in Docker

a1exsh:


a1exsh:
This is just a proof-of-concept.  If this considered to be OK, I'll add the doc changes.

femueller:
In general the PR looks fine to me, but could cause maybe later on issues if the Docker container UID changes. If that's fine I'm sure we can accept this pull request.

a1exsh:
@femueller thanks for looking into it.  What about instead `writable: true` we use this syntax `owner: application` and then we will just `chown` to whatever user is specified there by name, on every startup.  I guess that would make it more transparent.

a1exsh:
Actually, forget about this.  Looks like I need something else anyway :-p

PR #184, opened 2016-03-03
Format partition if mount point doesn't exist

### LINKED ISSUE #183: Partition not formatted on first boot
If I specify something like the following in the config YAML:

```
mounts:
  /var/lib/data:
    partition: /dev/xvdf
```

The volume would not be formatted and therefore could not be mounted.  In order to get it actually formatted one needs to specify `erase_on_boot: true`, but I don't want the data to be wiped on every EC2 restart (this might happen during normal lifetime of my instance).

Would it be possible to only format on first boot up?

a1exsh:
Also avoid hiding function dir_exists() with local variables of the same name.

Fixes #183


femueller:
Looks good to me

PR #197, opened 2016-03-22
Use first port as default health check port

twz123:
The [documentation][doc] says that the first port of `ports` is picked as default health check port. Remove the sorting to ensure that the implementation matches the documentation.

[doc]: https://stups.readthedocs.org/en/latest/components/taupage.html?highlight=first#health-check-path


hjacobs:
Hmm, that might break things for some people... :smirk:

twz123:
Yeah, may be the case. Would you suggest to change the docs instead?

hjacobs:
@twz123 yes, that's the safer bet I would say..

femueller:
I've removed `first` from the docs. (https://github.com/zalando-stups/documentation/commit/1ae9515d23d753948733e3e69337b53eb9d32c2e)

PR #198, opened 2016-03-23
Add vmstat to node exporter

csenol:


femueller:
Looks good to me

femueller:
This has to be reinvestigated. Apparently adding this option it breaks the tests. I've rolled this back.

csenol:
@femueller Is there another way to get context switches from taupage ?

femueller:
@csenol I'll look into it again. Maybe we just have to update the node exporter to make it work.

Issue #175, opened 2016-01-29
Ability to specify parameters (overcommit, shmmax)


feikesteenbergen:
The trigger for this issue is a spilo issue: https://github.com/zalando/spilo/issues/42

We would like to be able to configure some parameters for applications running on top of Taupage.
If we can specify it in some way in Docker it would be great, however I'm not sure that's possible.

The following parameters are normally tuned:
- `kernel.shmall`: physical_pages / 2
- `kernel.shmmax`: physical_pages \* pagesize / 2
- `vm.dirty_background_ratio`: 1
- `vm.dirty_ratio`: 2
- `vm.overcommit_memory`: 0
- `vm.swappiness`: 0

feikesteenbergen:
Could you provide some feedback on this issue? Do you want to implement this? We really need this.

hjacobs:
:+1: for the tuning, not sure how to implement it (configuration should be via Taupage properties)

PR #201, opened 2016-03-31
Allow certain sysctl parameters to be overridden

feikesteenbergen:
Tuning applications on top of Taupage sometimes requires tuning certain (kernel)
parameters. This allows Taupage configurations to have a sysctl section which
will be used instead of the defaults.
As some kernel parameters may not be allowed to be tuned, only parameters
on a whitelist are allowed to be specified.

This addresses issues:
* https://github.com/zalando-stups/taupage/issues/175
* https://github.com/zalando/spilo/issues/42

### DIFF
--- a/runtime/opt/taupage/init.d/06-update-sysctl.py
+++ b/runtime/opt/taupage/init.d/06-update-sysctl.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+
+import logging
+import sys
+import subprocess
+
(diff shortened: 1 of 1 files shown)


feikesteenbergen:
Could I get some feedback on this PR?

femueller:
Looks very good to me @feikesteenbergen 👍 .
However, this first needs to be properly integrated with ServerSpec. I'll have a look into it.

PR #207, opened 2016-04-12
Log the difference in taupage.yaml on startup

a1exsh:
For the stateful applications it might be helpful to have the changes
to Taupage configurtion in the EC2 instance user data to be logged
across restarts.

Using diff with zero lines of unified context: we are only interested
in the changed values.

### DIFF
--- a/runtime/usr/lib/python2.7/dist-packages/cloudinit/handlers/zalando_ami_config.py
+++ b/runtime/usr/lib/python2.7/dist-packages/cloudinit/handlers/zalando_ami_config.py
@@ -1,5 +1,8 @@
 # vi: ts=4 expandtab
 
+import shutil
+import subprocess
+
 from cloudinit import handlers
 from cloudinit import log as logging
 from cloudinit import util
@@ -12,6 +15,7 @@ TAUPAGE_AMI_CONFIG_PREFIX = "#taupage-ami-config"
 TAUPAGE_AMI_CONFIG_MIME_TYPE = handlers.type_from_starts_with(TAUPAGE_AMI_CONFIG_PREFIX)
 
 TAUPAGE_CONFIG = "/meta/taupage.yaml"
+TMP_TAUPAGE_CONFIG = "/tmp/taupage.yaml"
 
 
 class ZalandoAMIConfigPartHandler(handlers.Handler):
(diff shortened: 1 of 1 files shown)


hjacobs:
Hmm, that does not honor masking sensitive information :disappointed:

hjacobs:
i.e. it would log potentially sensitive information (e.g. think about Plan B Cassandra admin password) to the logging provider => not good

a1exsh:
This is sent to cloud-init-output.log. does this get uploaded to log
service also?

A password will be logged only if it was changed, and yes we can try to
sed/grep it.

hjacobs:
@a1exsh can you check syslog (which is pushed to log provider)?

a1exsh:
Nothing from that is logged to syslog.

The password will be logged to that could-init-output.log though on first bootup I think.

runtime/usr/lib/python2.7/dist-packages/cloudinit/handlers/zalando_ami_config.py
```diff
@@ -37,4 +41,10 @@ def handle_part(self, _data, ctype, filename, payload, frequency):
 
             LOG.debug("Storing merged configuration...")
             config_yaml = util.yaml_dumps(config_merged)
-            util.write_file(TAUPAGE_CONFIG, config_yaml, 0o444)
+            util.write_file(TMP_TAUPAGE_CONFIG, config_yaml, 0o444)
+
+            LOG.debug("Comparing current configuration with the old one...")
+            subprocess.call(['diff', '-u0', TAUPAGE_CONFIG, TMP_TAUPAGE_CONFIG])
```
hjacobs: this will fail the first time (/meta/taupage.yaml not there yet)? would be annoying to see an error in the logs..

a1exsh: Ah, yes, good point.  Will fix.

a1exsh: Actually, should it not be throwing an error earlier when it tries to read the config with:
```
            LOG.debug("Loading existing configuration...")
            config_yaml = util.read_file_or_url(TAUPAGE_CONFIG)
```
?

hjacobs: Ah yes, we always have a /meta/taupage.yaml containing default values (created during Taupage AMI build).

a1exsh: Good, then there's nothing to change at this line. :-)

femueller:
Looks good to me 👍

Issue #210, opened 2016-04-18
fix docker in docker support


ChristianLohmann:
bind mount any shared lib that is used by the docker client which is missing by default in the container.

Points to consider:
* what happens, if a newer image than ubuntu 14.04 LTS is used as container image? Will we overwrite existing libs?

femueller:
This issue can be closed, since previous PRs have been merged already.

PR #211, opened 2016-04-18
#210 bind mounting missing shared libs for docker to function properly

ChristianLohmann:
also works with docker 1.11.0


hjacobs:
Hard to see whether this will work (by looking at the bind mounts only..) --- I can only trust you to have it tested :smirk: 

Do we have a integration test in the Taupage build pipeline for this (should be easy to accomplish)?

ChristianLohmann:
@hjacobs i'm fully with you. I tested it locally with a plankton image. But having an integration test would be great.

hjacobs:
:+1:

PR #213, opened 2016-04-27
#210 changed plankton version to use

ChristianLohmann:


test-userdata.yaml
```diff
@@ -4,7 +4,7 @@ application_id: my-http-test-app-ami
 application_version: "1.1"
```
mrandi: version should be dynamic, something like that
```application_version: "{{Arguments.ImageVersion}}"```
(and be registered in yourturn)

File: test-userdata.yaml [hunk 2]
```diff
@@ -4,7 +4,7 @@ application_id: my-http-test-app-ami
...
 
 runtime: Docker
-source: registry.opensource.zalan.do/stups/plankton:cd13
+source: registry.opensource.zalan.do/stups/plankton:cd14
```
mrandi: This could also be dynamic:
```source: registry.opensource.zalan.do/stups/plankton:{{Arguments.ImageVersion}}```

ChristianLohmann:
adapted app name and version to the real plankton name and version

mrandi:
👍

PR #219, opened 2016-04-29
#218 use log sink

### LINKED ISSUE #218: use log-sink instead of fullstop to push logs
We created log-sink https://github.com/zalando-stups/logsink, migrating the instance-logs endpoint that was before in fullstop.
We need now to change endpoints and auth mechanism.

mrandi:
fixes #218

### DIFF
--- a/runtime/opt/taupage/bin/push-audit-logs.py
+++ b/runtime/opt/taupage/bin/push-audit-logs.py
@@ -11,11 +11,13 @@ import requests
 import sys
 import time
 
-from taupage import configure_logging, get_config, get_token, get_boot_time
+from base64 import b64encode
+
+from taupage import configure_logging, get_config, get_boot_time
 
 
 def push_audit_log(config: dict, instance_logs_url, account_id, region, instance_id, boot_time, fn, compress=False):
-    token = get_token(config, 'taupage', ['uid']) or {}
+    userAndPass = b64encode(b"test:test").decode("ascii") or ''
 
     with open(fn, 'rb') as fd:
         contents = fd.read()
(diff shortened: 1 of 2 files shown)


runtime/opt/taupage/bin/push-audit-logs.py
```diff
@@ -11,11 +11,13 @@
...
 
-from taupage import configure_logging, get_config, get_token, get_boot_time
+from base64 import b64encode
+
+from taupage import configure_logging, get_config, get_boot_time
 
 
 def push_audit_log(config: dict, instance_logs_url, account_id, region, instance_id, boot_time, fn, compress=False):
-    token = get_token(config, 'taupage', ['uid']) or {}
+    userAndPass = b64encode(bytes(config.get('logsink_username').append(':').append(config.get('logsink_password')), encoding='ascii')).decode("ascii") or ''
```
tuxlife: @mrandi Please split up this line

for example:
```python
    userAndPass = b64encode(bytes(':'.join(
            config.get('logsink_username'),
            config.get('logsink_password')
        ), encoding='ascii')).decode("ascii") or ''
```

mrandi: @tuxlife  already committed with append ;) is still ok?

tuxlife: @mrandi no, this is why i command your request.

tuxlife: @mrandi please use the format function for this.

and why log sink doesn't support oauth?

mrandi: because a lot of app, do not need any credentials. But we still need their logs.

PR #224, opened 2016-05-04
Start docker container if it already exists. Fix #199

### LINKED ISSUE #199: Docker doesn't start after EC2 instance reboot
The docker container cannot be started by Taupage after an EC2 instance reboot (from EC2 Management Console, or `sudo reboot` *or* auto-recovery).  The following error appears in the syslog:
```
Mar 31 09:10:10 ip-172-31-6-75 taupage-init: INFO: Docker run failed (try 2/3), retrying in 5s..
Mar 31 09:10:16 ip-172-31-6-75 taupage-init: Error response from daemon: Conflict. The name "taupageapp" is already in use by container 6012dd71e3bb. You have to remove (or rename) that container to be able to reuse that name.
```

This is a critical issue for planb-cassandra, as we rely on auto-recovery to bring the instances back up in case of system errors.

a1exsh:
This is a workaround for an issue with startup sequence when we
have external volumes to mount.  In case of EC2 instance restart
the Docker daemon will start prior to taupage-init and will
immediately try to start the container it was running prior to
the restart.  This will fail because the volumes are not there yet,
they are mounted as part of taupage-init script, in 10-prepare-disks.py.

Starting the container explicitly "fixes" the problem because at that
point all required initialization steps are completed.

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -304,7 +304,16 @@ def main(args):
         logging.error("Failed to list existing docker containers: %s", str(e))
         # not a fatal error, continue
 
-    if not already_exists:
+    if already_exists:
+        try:
+            cmd = ['docker', 'start', 'taupageapp']
+            logging.info('Starting existing Docker container: {}'.format(cmd))
+            if not args.dry_run:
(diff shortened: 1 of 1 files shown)


femueller:
Looks good to me 👍

PR #227, opened 2016-05-04
#226 move etcd log info

### LINKED ISSUE #226: move "register-in-etcd" log info from syslog to etcd.log
move "register-in-etcd" log info from syslog to etcd.log

mrandi:
fixes #226


femueller:
Looks good to me 👍

PR #230, opened 2016-05-20
#229 mount var/log as read-write

### LINKED ISSUE #229: mount log directory as read-write in order to write custom log files
```
# mount logdirectory as read-write in order to write custom log files
    if config.get('mount_var_log'):
        yield '-v'
        yield '/var/log:/var/log'
```

mrandi:
fixes #229


apfeiffer85:
:+1:

tuxlife:
:-1:

tuxlife:
With this fix is it possible to change and remove the taupage systemlogs

apfeiffer85:
ok so we have to use some different directory for custom logs like /var/log/custom or something similar.

apfeiffer85:
we now only mount /var/log/custom as read-write

mrandi:
👍

tuxlife:
@apfeiffer85  Why did you need /var/log from the host under /var/log in docker?

Why not /var/log-docker/:/var/log and /var/log:/var/log-host?

apfeiffer85:
i guess there are some use cases where Teams need read access to /var/log/application.log and maybe other log files therefore was the read-only mount (this was already implemented). 
And i guess there is no really need to mount /var/log into /var/log in the docker container it just was because of common place for logs. 
We could definitely change this to something else like /var/log-host or anything else.

But what do you mean with /var/log-docker:/var/log ? 

What we also did was to add a custom logdir /var/log/custom for custom application logs so we only mount /var/log/custom as read-write. But we also could mount this into /var/log-host/custom . This is necessary because the log shipping agents running on the host system and they can only follow logs from the host system ( or lets say it is easier to follow logs on the host system ).

tuxlife:
:+1:

apfeiffer85:
New Solution is now: 

mount /var/log from host system to the container dir /var/log-host (read-only)
And any custom logs will be saved at /var/log-custom (host system) and mounted as read-write to the container /var/log

PR #231, opened 2016-05-20
install xenial kernel 4.4.0-22-generic

mfelsche:
as using the stock 14.04 kernel (3.13.X) is preventing sshd to work inside the docker image
after upgrading sshd.

example error output:

nor calling  with --capabilities-add AUDIT_WRITE nor --privileged=true resolved this.
A kernel upgrade on taupage and a reboot did help.


tuxlife:
@mfelsche Please describe your Issue. We use ssh in a docker image with our [Odd](https://github.com/zalando-stups/odd) and it works without problems.

hjacobs:
Yes, please let's not just install a different Kernel because somebody mentions "problems" without describing details.

mfelsche:
i will provide a small reproducable example, gimme a sec

szuecs:
@mfelsche please create an Issue first, such that we can discuss it there. Thanks!

Issue #237, opened 2016-06-02
Include necessary Scala/Play/akka classes in AppDynamics JVM-Agent


apfeiffer85:
This should be configurable with the taupage.yaml . So per default all calsses are excluded but there should be an option to include these classes. 

                    <excludes filter-type="STARTSWITH" filter-value="scala.concurrent/"/>
                    <excludes filter-type="STARTSWITH" filter-value="akka/"/>
                    <excludes filter-type="STARTSWITH" filter-value="play.core.server/"/>
                    <excludes filter-type="STARTSWITH" filter-value="play.api.libs.concurrent/"/>

PR #238, opened 2016-06-02
add option to include necessary scala classes in AppDynamics 

apfeiffer85:
issue #237


Raffo:
@apfeiffer85 Can you explain a bit more the changes? I don't have a way to prove they are right and so I cannot merge the PR. Also, I saw the comment by @hjacobs on the documentation, we can maybe also address it here where the PR lives.

apfeiffer85:
@Raffo the PR is about to change the default configuration from the AppDynamics JVM Agent. And i also answer to the question in the documentation. Unfortunately there is no better way to do this. :( 
And this must be optional because if we change this to every single JVM Agent these can cause performances issues. So only Scala/play/akka Applications need this modifications.

Raffo:
Well, it's good then.

Raffo:
LGTM

PR #239, opened 2016-06-06
Disable THP and add add net.core.somaxconn to sysctl whitelist

CVirus:


CVirus:
@feikesteenbergen take a look if you have the time please

Raffo:
@CVirus Can you add here the issue why you need this PR? Thx.

CVirus:
@Raffo This is needed to be able to get redis running properly on docker .. If you try to run redis; you'll get the following:

`2914:M 07 Jun 11:48:59.324 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
2914:M 07 Jun 11:48:59.324 # Server started, Redis version 3.2.0
2914:M 07 Jun 11:48:59.324 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
2914:M 07 Jun 11:48:59.324 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
2914:M 07 Jun 11:48:59.324 * DB loaded from disk: 0.000 seconds
2914:M 07 Jun 11:48:59.325 * The server is now ready to accept connections on port 6379`

szuecs:
@CVirus Do you have a Github Issue in the taupage repository?
I know that redis and Postgres have requirements to change kernel setting using sysctl.

CVirus:
No, I'm sorry I don't.. Do you need me to create one ?

szuecs:
@CVirus It would be very nice, if you would do. 
Thanks and all the best, sandor

CVirus:
@szuecs @rafaelcaricio #242 #241

szuecs:
@CVirus Please split the PR into, two, one for each issue.
#241 is mergable, but #242 is not.

The problem that I see is that the fix for #242 is to specific. You could do it better if you would have similar to sysctl a whitelist which paths in /sys and /proc you are allowed to change and then a function that use the yaml structure to echo the data specified there into the specified and whitelisted file.

Best and thanks, sandor

CVirus:
@szuecs #243 is a fix for #241 .. Will work on a fix for #242 during this week .. Thanks!

PR #244, opened 2016-06-14
Support for editing sysfs paths

### LINKED ISSUE #242: Support disabling THP in the kernel
Some apps (like redis) require Transparent Huge Pages (THP) support to be disabled in the kernel, as shown in the following warning msg from redis:

`# WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.`

CVirus:
Should fix #242


szuecs:
lgtm

PR #247, opened 2016-06-24
disable follow AppD logfiles per default

apfeiffer85:
I want to disable the AppDynamics log parser per default. So the AppD Agents logs will not be pushed to the AppDynamics Analytics cluster anymore until you especially activate it in the JobFiles.


apfeiffer85:
you can close this PR as it is also included in my last PR: https://github.com/zalando-stups/taupage/pull/250

PR #251, opened 2016-07-11
enabling appdynamics machine agent http listener

danmanirl:
I would like to send custom metrics to the machine-agent http listener as described in:

https://docs.appdynamics.com/display/PRO41/Standalone+Machine+Agent+HTTP+Listener

Is there a possibility of adding this option by default to `appdynamics.conf` ?


apfeiffer85:
You have to set the host, so please add:  ‑Dmetric.http.listener.host=0.0.0.0

apfeiffer85:
👍

PR #261, opened 2016-08-09
add applicationname to controller-info.xml if machine agent start wit…

apfeiffer85:
we need this to run the machine agent as standalone process without the AppAgent


astarnell:
:+1:

Issue #252, opened 2016-07-15
Resize volume if filesystem has free space


feikesteenbergen:
Recently, when running a Spilo with a larger sizes EBS volume than the snapshot it was based upon, it required a manual `resize2fs`

Proposal:
- If the mounted filesystem does not need to be formatted
- If the formatted filesystem is smaller than the volume
- Then resize2fs

slitsche:
+1
We also need this for scaling up planb-cassandra instances.

feikesteenbergen:
Feature is merged.

PR #264, opened 2016-08-11
Resize mounts by default

feikesteenbergen:
Volumes that should not be erased_on_boot may have been created
from different snapshots. One of the reasons to do this is to be able
to scale data stores (PostgreSQL, Cassandra) once they fill up.
What happens in this case is that the volume is resized, however the
filesystem still has the previous size.
Resizing the filesystem to the volume seems like a win, we can do
it unconditionally as resize2fs will figure out by itself whether
a resize is needed.

Addresses issue #252

### DIFF
--- a/runtime/opt/taupage/init.d/10-prepare-disks.py
+++ b/runtime/opt/taupage/init.d/10-prepare-disks.py
@@ -125,6 +125,9 @@ def iterate_mounts(config):
         def mount():
             mount_partition(partition, mountpoint, options, filesystem, os.path.isdir(mountpoint), already_mounted)
 
+        def resize():
+            subprocess.check_call(['resize2fs', partition])
+
         if partition and not already_mounted:
             if initialize:
                 format()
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/init.d/10-prepare-disks.py
```diff
@@ -125,6 +125,9 @@ def format():
         def mount():
             mount_partition(partition, mountpoint, options, filesystem, os.path.isdir(mountpoint), already_mounted)
 
+        def resize():
+            subprocess.check_call(['resize2fs', partition])
```
slitsche: What if it the file system is not ext4? Shouldn't we call `xfs_growfs` for xfs?

feikesteenbergen: Yes, we need that

szuecs:
:+1:

szuecs:
:+1:

PR #270, opened 2016-08-18
remove whitelist for setting sysctl values

szuecs:
There is no reason to have a whitelist, because settings are tracked in git senza repos.
To tune contrack table timeouts we need to tune socket parameters here.


tuxlife:
:+1:

astarnell:
:+1:

PR #288, opened 2016-08-23
add local planb-tokeninfo

szuecs:
Add local planb-tokeninfo to remove the rtt to an external tokeninfo service.

### DIFF
--- a/runtime/etc/init/planb-tokeninfo.conf
+++ b/runtime/etc/init/planb-tokeninfo.conf
@@ -1,0 +1,9 @@
+description "PlanB Tokeninfo"
+
+stop on shutdown
+
+respawn
+
+script
+	build/planb-tokeninfo
+end script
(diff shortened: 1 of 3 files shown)


tuxlife:
:+1:

runtime/etc/init/planb-tokeninfo.conf
```diff
@@ -0,0 +1,9 @@
+description "PlanB Tokeninfo"
+
+stop on shutdown
+
+respawn
+
+script
+	build/planb-tokeninfo
```
hjacobs: is "build" the right path? I guess not?

lukasniemeier-zalando:
Cool, is this already released?

szuecs:
@lukasniemeier-zalando no, but we plan to release it today

hjacobs:
This does not work as expected yet --- we should also set the `TOKENINFO_URL` env var to "http://localhost:9021/oauth2/tokeninfo" for the application Docker container.

szuecs:
@hjacobs why? If I understand it correctly the user can and should define it herself.

hjacobs:
@szuecs we set `TOKENINFO_URL` in Taupage, so we should automatically set it to the correct localhost URL when starting it locally. The user should not be aware on which port it runs locally...

szuecs:
@hjacobs I will set it in our Zalando specific configuration, otherwise it would not make sense.

PR #289, opened 2016-08-23
pass variables from build config to runtime if available and pass the…

szuecs:
…m to planb-tokeninfo

run planb-tokeninfo as nobody


tuxlife:
:+1:

PR #293, opened 2016-08-29
Changed the hourly log rotation call 

### LINKED ISSUE #292: Add "nice" and "ionice" to logrotate
Right now logrotate on bigger log volumes can kill performance during compression and rotation phase.

Taupage should use something like this (from google):

```
nice -n 19 ionice -c3 /usr/sbin/logrotate /etc/logrotate.conf
```

lmineiro:
This should be nicer on cpu and i/o scheduling.

Closes #292


szuecs:
:+1:

Issue #296, opened 2016-08-30
Push audit-logs with a randomized offset


ChristianLohmann:
Actually every Taupage instance pushes its rotated audit logs every 17 minutes after a full hour. This results in a massive increase of requests/s on log-sink. 
To avoid those hourly peaks it would be better to push the audit-logs after a randomized offset.
The easiest way would be to implement it in the push-audit-logs.py file

PR #297, opened 2016-08-30
push-audit-logs.py should spread the requests to audit API #296

szuecs:
Used a random integer between 60 and 3500 seconds to send data, after the first of the script run, such that on shutdown it will processed without sleep.


harti2006:
:+1:

szuecs:
@harti2006 I need again an approval

harti2006:
:+1:

PR #298, opened 2016-08-30
Change the source of TOKENINFO_URL

szuecs:
Use LOCAL_TOKENINFO_URL from our deployment config


szuecs:
@hjacobs this would address your suggestion to set it per default

hjacobs:
I don't understand this change --- I meant changing the `TOKENINFO_URL` to http://localhost:9021 for the application Docker container, here you change it for the planb-tokeninfo binary.

PR #301, opened 2016-09-05
Set TOKENINFO_URL to localhost:9021 if Plan B Token Info runs locally

hjacobs:
Set `TOKENINFO_URL` to "http://localhost:9021/oauth2/tokeninfo" if "local_planb_tokeninfo" is set.


hjacobs:
:+1:

lmineiro:
👍

PR #312, opened 2016-09-29
bugfix: script was using wrong variable to decrypt ACCESSKEY. fix #311

### LINKED ISSUE #311: Appdynamic startup with encrypted password fails
We set the appdynamics_account_access_key with aws:kms:MyEncryptedKey in the senza file and the startup of the application fails.

We found the issue in:
/runtime/opt/taupage/init.d/85-setup-appdynamics.sh Line 14-18

```
if [[ $ACCESSKEY == "aws:kms:"* ]]; then
	ACCOUNTKEY=${ACCESSKEY##aws:kms:}
	ACCOUNTKEY=`python3 /opt/taupage/bin/decrypt-kms.py $ACCESSKEY`
	#overwrite ACCESSKEY with decrypted value
	ACCESSKEY=$ACCOUNTKEY
fi
```
ACCOUNTKEY contains the value without aws:kms: but
decrypt-kms.py is called with ACCESSKEY which still contains aws:kms:
decrypt-kms.py expects the key without aws:kms:

apfeiffer85:


tuxlife:
:+1:

PR #316, opened 2016-10-13
mount node-js AppDynamics directory with snippet file to the Docker C…

### LINKED ISSUE #315: Mount AppDynamics node.js Agent-Snippet Directory to Docker
There is a small snippet file with important values for AppDynamics node.js instrumentation / implementation. 
Currently this is only available on the Hostsytem. We need to mount that also to the Docker container.

apfeiffer85:
…ontainer - fix #315


tuxlife:
:+1:

PR #317, opened 2016-10-19
truncate the nodename if it has more than 100 chars - fix #314

### LINKED ISSUE #314: AppDynamics does not support nodenames with more as 100 chars.
AppDynamics does not allow nodenames with more as 100 Chars. Therefore we have to build a truncate solution. 

Every nodename which is longer as 100 chars will be truncated.

apfeiffer85:


runtime/opt/taupage/init.d/85-setup-appdynamics.sh
```diff
@@ -50,7 +50,9 @@ if [ -n "$ACCOUNT_NAME" ]; then
 fi
 
 # node name has to be unique across the whole ecosystem
-node="${config_notify_cfn_stack}_$(hostname)_$(ec2metadata --availability-zone)_$(ec2metadata --instance-id)"
+generated_nodename="${config_notify_cfn_stack}_$(hostname)_$(ec2metadata --availability-zone)_$(ec2metadata --instance-id)"
+#truncate the nodename if it has more than 100 chars
+node=$(echo $generated_nodename | cut -c 1-99)
```
tuxlife: @apfeiffer85 Please use Substring Expansion instead of a subshell with echo and cut
```
$ foo=abcdefghijklmnopqrstuvwxyz
$ echo ${foo:0:5}
abcde
```

apfeiffer85: good idea, will do that.

tuxlife:
:+1:

PR #327, opened 2016-12-01
#318 blow up on human error (wrong type in YAML)

### LINKED ISSUE #318: Volume mount erase_on_boot flag interprets non-empty strings as true by default
When deploying a stack backed by ebs volume using senza, the senza.yaml config interprets a non-empty string in `erase_on_boot` flag as true. For instance the following config 

```
mounts:
   /data:
      partition: /dev/xvdk
      filesystem: ext4
      erase_on_boot: "false"
      options: noatime,nodiratime,nobarrier
```

has resulted in our persisted data being wiped on every reboot. It would be great if there could be some check for this since it is very easy to miss and hard to troubleshoot.

hjacobs:
Fixes #318.

Blow up if the user mistakenly enters a string in the user data YAML for `erase_on_boot` parameter.

This will prevent "frequent" errors like putting the string "false" into the YAML (non-empty string evaluates to `True`).

### DIFF
--- a/runtime/opt/taupage/init.d/10-prepare-disks.py
+++ b/runtime/opt/taupage/init.d/10-prepare-disks.py
@@ -142,6 +142,9 @@ def iterate_mounts(config):
         partition = data.get("partition")
         filesystem = data.get("filesystem", "ext4")
         initialize = data.get("erase_on_boot", False)
+        if not isinstance(initialize, bool):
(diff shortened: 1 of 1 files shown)


a1exsh:
:+1:

a1exsh:
Thanks, @hjacobs!

retnuh:
Thanks!

PR #328, opened 2016-12-01
bugfix appd scala include classes - fix #326

### LINKED ISSUE #326: appdynamics_include_scala_classes: true flag does not seem to work
Hi,
tried this flag on the taupage release `Taupage-AMI-20161109-140716 ami-d3f70dbc` but it does not seem to work.

**[/opt/proprietary/appdynamics-jvm/latest_version/conf/app-agent-config.xml](https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/85-setup-appdynamics.sh#L164)** still has:
```
<configuration>
                <fork-config>
                    <!-- Scala/Akka/Play  -->
                    <excludes filter-type="STARTSWITH" filter-value="scala.concurrent/"/>
                    <excludes filter-type="STARTSWITH" filter-value="akka/"/>
                    <excludes filter-type="STARTSWITH" filter-value="play.core.server/"/>
                    <excludes filter-type="STARTSWITH" filter-value="play.api.libs.concurrent/"/>
...
```

our .yml file:
```
TaupageConfig:
  appdynamics_application: "Brand Analytics"
  appdynamics_include_scala_classes: true
```

apfeiffer85:


szuecs:
:+1

PR #341, opened 2016-12-14
fix know kernel issues if you have heavy write load like PG

szuecs:
I hope it is possible to update to a newer Kernel. Please share with our DB team


Raffo:
👍

PR #342, opened 2016-12-14
Rework filesystem initialization logic

a1exsh:
Resize function doesn't need a special check for e2fsck as we are
running the check ourselves prior to mount.  Also renamed `resize` to
`extend`.


szuecs:
:+1:

PR #344, opened 2016-12-15
we want to see everything in STDOUT and not search in files

szuecs:
The log in our build pipeline does not show the content of log files. We are not using these files anyways so delete it


mikkeloscar:
:+1:

PR #345, opened 2016-12-15
Bugfix/do not call external sites

tuxlife:


hjacobs:
What do we need the yandex key for?

tuxlife:
we don't need the yandex key. it was the docker key

szuecs:
:+1

PR #347, opened 2016-12-15
add yaml-dev for building python yaml dependency with pip

szuecs:
While this does not break the build, it's a bug that should be fixed


tuxlife:
:+1:

PR #350, opened 2016-12-16
Fix call to e2fsck: allow to run non-interactively

a1exsh:


a1exsh:
Actually, that's still not good enough.  We need to examine the exit code of `e2fsck` carefully, according to the man page:
```
EXIT CODE
       The exit code returned by e2fsck is the sum of the following conditions:
            0    - No errors
            1    - File system errors corrected
            2    - File system errors corrected, system should
                   be rebooted
            4    - File system errors left uncorrected
            8    - Operational error
            16   - Usage or syntax error
            32   - E2fsck canceled by user request
            128  - Shared library error
```

runtime/opt/taupage/init.d/10-prepare-disks.py
```diff
@@ -105,16 +115,33 @@ def format_partition(partition, filesystem="ext4", initialize=False, is_already_
...
+        except CmdException as e:
+            # see e2fsck(8) man page for description of exit codes
+            if e.returncode <= 1:
+                # all OK
+                return
+            elif (e.returncode & 2) != 0:
+                logging.exception("File system errors corrected on %s, system should be rebooted.", partition)
+            elif (e.returncode & 4) != 0:
+                logging.exception("File system errors left uncorrected on %s.", partition)
+            elif e.returncode == 8:
```
szuecs: We already had an issue with exitcode 8 from e2fsck, which erased all our pipelines. 
I will change this to logging.exception, before merge.

szuecs:
:+1:

runtime/opt/taupage/init.d/10-prepare-disks.py
a1exsh: If we don't raise an exception here, then a genuinely empty volume will never be formatted at boot time (unless `earase_on_boot` is set). For our use case we want it formatted on first boot, but never on subsequent. Unfortunately, there doesn't seem to be a reliable way to detect first boot, so we have to resort to such ugly and risky tricks... :-(

PR #351, opened 2016-12-19
Do not format for e2fsck with exitcode 8

szuecs:
Do not format for e2fsck with exitcode 8, because our Go-CD pipeline server was deleted, because of that


Raffo:
👍

Issue #354, opened 2017-01-02
the /var/log/etcd.log file is not rotated


mrandi:
etcd.log file is not rotated

```5.1G -rw-r-----  1 syslog    adm    5.1G Jan  2 09:08 etcd.log```

szuecs:
Please make reasonable rotation, maybe by size and time and let some files on disk

mrandi:
@szuecs commented on the pull request

PR #355, opened 2017-01-02
rotate etcd logs

mrandi:
fixes https://github.com/zalando-stups/taupage/issues/354

### DIFF
--- a/runtime/etc/logrotate.d/etcd
+++ b/runtime/etc/logrotate.d/etcd
@@ -1,0 +1,11 @@
+/var/log/etcd.log
+{
+        rotate 1
+        hourly
+        missingok
+        notifempty
+        copytruncate
+        compress
+        delaycompress
+        sharedscripts
+}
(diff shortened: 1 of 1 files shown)


mrandi:
👍

ehartung:
:+1:

runtime/etc/logrotate.d/etcd
```diff
@@ -0,0 +1,11 @@
+/var/log/etcd.log
+{
+        rotate 1
```
szuecs: Why you only let 1 file and rotate every hour?

szuecs:
:-1:

runtime/etc/logrotate.d/etcd
mrandi: @szuecs What is the problem with this configuration? What would be your suggestion?
I always need the latest log on the machine, for older entries I can take a look also in other places.

szuecs: @mrandi Think about other people which may rely on having these logs.
My suggestion is at least have files for one week.

mrandi:
👍

szuecs:
:+1:

PR #356, opened 2017-01-05
Format EBS volumes based on Tag and User Data

a1exsh:
This fixes the way we decide if a volume needs to be formatted or not.
We will only format an EBS volume if it is required explicitly by the
erase_on_boot flag in the user data or if a special temporary tag is set
on the volume.  We remove such tag immediately when formatting, so that
the next time instance starts up it will not try to format again in
error.

This makes the use case of RAID volumes over multiple EBS devices not
supported unless erase_on_boot is set to True in the user data.


Raffo:
👍

PR #358, opened 2017-01-05
Fix use of EC2 client in prepare disks script

a1exsh:
Only require a client when actually needed, avoid doing that on script
startup: instances running without any role assigned will just have an
exception there.

### DIFF
--- a/runtime/opt/taupage/init.d/10-prepare-disks.py
+++ b/runtime/opt/taupage/init.d/10-prepare-disks.py
@@ -14,6 +14,9 @@ from time import sleep
 from taupage import configure_logging, get_config
 
 
+current_region = None
+
+
 def instance_id():
     """Helper to return theid for the current instance"""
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/init.d/10-prepare-disks.py
```diff
@@ -372,17 +386,17 @@ def main():
     else:
         configure_logging(logging.INFO)
 
+    global current_region
```
Raffo: Why do we need this? Using `global` smells a bit to me, but I'm not really a python expert.

a1exsh: That is just a short cut since region can be overridden from command line and the args are only available in `main()`.  We'll try to come up with a cleaner workaround.

a1exsh:
@Raffo please review the updated version :)

Raffo:
👍

PR #362, opened 2017-01-18
mount AppDynamics MachineAgent extension Directory to the Docker 

apfeiffer85:
Then the Developer is able to add custom AppDynamics MachineAgent extensions.


szuecs:
:+1

PR #232, opened 2016-05-30
Add NVIDIA driver and CUDA support to Taupage [work in progress]

elezar:
Closes #240.

This PR adds the NVIDIA driver (and CUDA support) to Taupage. This should allow for GPU instances to be used for compute intensive applications such as machine learning, while still maintaining the audit compliance provided by Taupage.

Note that the installation of CUDA is delegated to the docker images, and thus the only overhead added to the AMI iteself is the NVIDIA driver. This means that additional space requirements are reduced. What are the limitations for STUPS AMIs?
- Image generation: 
  - [x] Install the NVIDIA driver.
- Runtime (when running on a g2.2xlarge): 
  - [x] Ensure that relevant device nodes are created.
  - [x] Modify the docker launch parameters to map devices and drivers to docker container.
- Testing:
  - [ ] Create a simple test GPU-based docker image to test the driver installation.
  - [ ] Ensure that the driver does not interfere on instances where no GPU devices are present.

Note that `g2.2xlarge` instances are required to run GPU applications on AWS. I must still confirm whether this is a strict requirement when building the AMI.

References:
- NVIDIA docker discussion on installing drivers on AWS: https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
- Arguments for mapping drivers and devices to docker containers: https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/docker
- Creating device nodes when X11 is not present: http://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#runfile-verifications

Issue #240, opened 2016-06-08
Add support for GPU-based instances


elezar:
For machine learning applications, it would be useful to have access to the AWS GPU instances.

Note that the installation of CUDA is delegated to the docker images, and thus the only overhead added to the AMI iteself is the NVIDIA driver. This means that additional space requirements are reduced. What are the limitations for STUPS AMIs?
- Image generation: 
  - Install the NVIDIA driver.
- Runtime (when running on a g2.2xlarge): 
  -  Ensure that relevant device nodes are created.
  -  Modify the docker launch parameters to map devices and drivers to docker container.
- Testing:
  - Create a simple test GPU-based docker image to test the driver installation.
  - Ensure that the driver does not interfere on instances where no GPU devices are present.

Note that `g2.2xlarge` instances are required to run GPU applications on AWS. I must still confirm whether this is a strict requirement when building the AMI.

References:
- NVIDIA docker discussion on installing drivers on AWS: https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
- Arguments for mapping drivers and devices to docker containers: https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/docker
- Creating device nodes when X11 is not present: http://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#runfile-verifications

fatz:
@elezar, @jbellmann guess this can be closed now

elezar:
Yes. Thanks @fatz. Closing this as it was addressed in #365

PR #365, opened 2017-02-02
WIP added nvdia drivers and docker plugin

fatz:
this is an alternative version of #232 without adding the dependencies of the nvidia packages.

In addition I added nvidias docker plugin.

Currently work in progress is the testing:
- [ ] Test cuda driver
- [ ] Test different instance types ( not sure how to implement this )
- [ ] Test Docker.py changes

### DIFF
--- a/build/setup.d/10-upgrade-base.sh
+++ b/build/setup.d/10-upgrade-base.sh
@@ -8,7 +8,8 @@ echo "Updating system..."
 apt-get update -y  # -q >>/tmp/build/upgrade.log
 
 # install 3.16. LTS kernel and make sure it updates to the last version
-apt-get install -y linux-image-virtual-lts-utopic
+# we need the general kernel for GPU instances
+apt-get install -y linux-image-generic-lts-utopic
 
 apt-mark hold openssh-server
 apt-get install -y --only-upgrade libc6 libssl1.0.0
--- a/build/setup.d/40-nvidia.sh
+++ b/build/setup.d/40-nvidia.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
+# But we're using the run script to not depend on build tools
+DRIVER_ARCH="Linux-x86_64"
+DRIVER_VERSION="375.26"
+DRIVER_FILENAME="NVIDIA-${DRIVER_ARCH}-${DRIVER_VERSION}.run"
+
+DOCKER_DRIVER_VERSION="1.0.0"
+DOCKER_DRIVER_FILENAME="nvidia-docker_${DOCKER_DRIVER_VERSION}-1_amd64.deb"
+
+
+apt-get update
+
+pkgs="
+build-essential
+linux-headers-generic-lts-utopic
+"
+
+#get the version of the lts kernel
+kernelname=$(apt-cache showpkg linux-headers-generic-lts-utopic | grep -a1 "Dependencies:" | tail -n1 | sed 's/.*linux-headers-\([0-9a-z.-]\+\).*/\1/')
+
+apt-get install -y -q --no-install-recommends -o Dpkg::Options::="--force-confold" $pkgs >>install.log
+
+echo "Installing NVIDIA driver ver: ${DRIVER_VERSION}"
+
+# Using the run File
+wget -P /tmp "http://us.download.nvidia.com/XFree86/${DRIVER_ARCH}/${DRIVER_VERSION}/${DRIVER_FILENAME}"
+
+chmod +x "/tmp/${DRIVER_FILENAME}"
+
+/tmp/$DRIVER_FILENAME -e -a -s --kernel-source-path "/usr/src/linux-headers-${kernelname}/" --kernel-name "${kernelname}"
+
+echo "Installing nvidia docker ver: ${DOCKER_DRIVER_VERSION}"
+wget -P /tmp "https://github.com/NVIDIA/nvidia-docker/releases/download/v${DOCKER_DRIVER_VERSION}/${DOCKER_DRIVER_FILENAME}"
+dpkg -i "/tmp/${DOCKER_DRIVER_FILENAME}"
+
+echo "Removing build packages driver"
+apt-get purge -y -q  $pkgs >>install.log
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -94,6 +94,13 @@ def wait_for_local_planb_tokeninfo():
     # => use global one
     return None
 
+def get_docker_command(config: dict):
+    '''get the docker command to use'''
+    cuda_device_files = glob.glob('/dev/nvidia*')
+    if size(cuda_device_files) > 0:
+        return 'nvidia-docker'
+    else:
+        return 'docker'
 
 def get_env_options(config: dict):
     '''build Docker environment options'''
@@ -345,9 +352,11 @@ def main(args):
 
     source = config['source']
 
+    docker_cmd = get_docker_command(config)
+
     already_exists = False
     try:
-        cmd = ['docker', 'ps', '-a', '-q', '-f', 'name=taupageapp']
+        cmd = [docker_cmd, 'ps', '-a', '-q', '-f', 'name=taupageapp']
         if subprocess.check_output(cmd):
             already_exists = True
     except Exception as e:
@@ -356,7 +365,7 @@ def main(args):
 
     if already_exists:
         try:
-            cmd = ['docker', 'start', 'taupageapp']
+            cmd = [docker_cmd, 'start', 'taupageapp']
             logging.info('Starting existing Docker container: {}'.format(cmd))
             if not args.dry_run:
                 subprocess.check_output(cmd)
@@ -369,7 +378,7 @@ def main(args):
         if registry:
             registry_login(config, registry)
 
-        cmd = ['docker', 'run', '-d', '--log-driver=syslog', '--name=taupageapp', '--restart=on-failure:10']
+        cmd = [docker_cmd, 'run', '-d', '--log-driver=syslog', '--name=taupageapp', '--restart=on-failure:10']
         for f in get_env_options, get_volume_options, get_port_options, get_other_options:
             cmd += list(f(config))
         cmd += [source]


runtime/opt/taupage/runtime/Docker.py
```diff
@@ -95,6 +96,15 @@ def wait_for_local_planb_tokeninfo():
     return None
 
 
+def get_docker_command(config: dict):
+    '''get the docker command to use'''
+    cuda_device_files = glob.glob('/dev/nvidia*')
```
elezar: Would it make sense to list the exact requirements here?
```
/dev/nvidia[0-9]+  
/dev/nvidiactl
/dev/nvidia-uvm
```
Although, I'm happy with the current implementation as a first step.

Does adding some logging for debugging startup make sense?

build/setup.d/40-nvidia.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
+# But we're using the run script to not depend on build tools
+DRIVER_ARCH="Linux-x86_64"
+DRIVER_VERSION="375.26"
+DRIVER_FILENAME="NVIDIA-${DRIVER_ARCH}-${DRIVER_VERSION}.run"
```
elezar: Do we want to include hash checks for the downloaded files?

Raffo:
Can we refer to issue #240 as requirements? If not, can you tell us more about what are you exactly trying to reach here? Also, happy to close the other one if you decide to go forward with this one only.

build/setup.d/40-nvidia.sh
fatz: thats in fact a good idea.

fatz:
Yes indeed its nearly the same requirement. 

But #240 uses the nvidia provided .dep packages which add a lot of build dependencies that can't be cleaned up anymore. This pull-req. has the approach to use nvidias build script instead. 

About closing #240: @elezar should decide. Currently i'm using the branch as a reference to not miss something which @elezar already did. But wanted to have a clean start.

So my suggestion is to wait 'till i'm finished with the tests and afterwards sort #240 out

elezar:
I have closed #232 as I think @fatz's implementation here is a little cleaner.

The issue #240 can still act as the specification, as it seems to be independent of the actual implementation details.

tests/spec/localhost/nvidia_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe command('nvidia-docker run --rm nvidia/cuda nvidia-smi') do
```
elezar: using `nvidia-docker run --rm nvidia/cuda nvidia-smi -L` should reduce the length of the output if no further checks are added.

elezar: An alternative would be to include a check for the driver version below.

fatz: yes this implementation is leaving out the fact that we might build on an instance without nvidia device.

but i really like the cuda containers `nvidia-smi` cause it handles a lot tests in one... but i'm open for suggestions

fatz:
ok just a small update for this. I'll try to do some progress on this PR this week.

Good news is that for the driver build we do not need an attached GPU. So we can build on any non p2 instance ( which currently aren't available in eu-central-1 ).

fatz:
Ok finally..

sorry for the delay. We now verified that our model training is working on this machines. 

Tests and build works for p2 and all other machines. Also the build has not to be started with a p2 instance. 

Integration tests cover p2 and not p2 instances by checking for nvidia device files

build/setup.d/40-nvidia.sh
```diff
@@ -0,0 +1,47 @@
...
+DRIVER_CHECKSUM="d60819b2e377398c7296999ab5e7c1a4"
+
+DOCKER_DRIVER_VERSION="1.0.0"
```
elezar: I see there is a 1.1.0 version out that mentions some bugfixes: https://github.com/NVIDIA/nvidia-docker/releases

Should we upgrade to it here too?

tests/spec/localhost/nvidia_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+has_nvidiactl = file('/dev/nvidiactl').exists?
+
+# should always be installed
+describe file('/usr/bin/nvidia-docker') do
+  it { should exist }
+end
+
+# we just can test this on gpu enabled instances.
+describe command('nvidia-docker run --rm nvidia/cuda nvidia-smi'), :if => has_nvidiactl do
+  its(:stdout) { should contain /Tesla K80/ }
```
elezar: Should we also check the output on non-GPU instances?

build/setup.d/40-nvidia.sh
fatz: indeed we can test this aswell...

tests/spec/localhost/nvidia_spec.rb
fatz: this would lead to errors 'cause the nvidia-docker expected devices are not in place. Thats the reason why i exclude it. 

However but i check if the command is installed. 

It might make sense to add an optional step in the `test.sh` for checking GPU instances. But there are no p2 instances in eu-central-1 so doing it all the time or exclusively adds additional constraints to the testing

fatz: @elezar you're right nvidia-docker still runs without gpus. So we should be able to test. But if you try the usual `docker run --rm nvidia/cuda nvidia-smi` it ends up in a not found...  However IMO it should be enough to have a test for the nvidia-docker executable + having a conditional check on GPU instances.

build/setup.d/40-nvidia.sh
fatz: have to rebase anyways... so working on that now

tests/spec/localhost/nvidia_spec.rb
elezar: Yes, you're right. Thinking about it again, this would add too much complexity without adding any additional value. Let's leave it like it is.

elezar:
Looks good to me. Thanks.

@szuecs @Raffo @tuxlife, any comments from your side?

build/setup.d/10-upgrade-base.sh
```diff
@@ -54,7 +54,7 @@ popd
 
 # install 3.16. LTS kernel and make sure it updates to the last version
 # also this step should build ixgbevf and ena kernel modules and put them into initramfs
-apt-get install -y linux-image-virtual-lts-xenial
+apt-get install -y linux-image-generic-lts-xenial
```
szuecs: Why this change is required?
We want to have auto updated kernels on rebuild

build/setup.d/20-install-packages.sh
```diff
@@ -18,7 +18,6 @@ libswitch-perl
 libwww-perl
 libyaml-dev
 libyaml-0-2
-linux-image-extra-virtual-lts-xenial
```
szuecs: Why do you drop it?
AFAIK it's required by other people.

build/setup.d/20-install-packages.sh
```diff
@@ -6,7 +6,7 @@
 pkgs="
 auditd
 build-essential
-docker-engine=1.11.1-0~trusty
+docker-engine=1.12.6-0~ubuntu-trusty
```
szuecs: Do you need a new feature from docker?

build/setup.d/10-upgrade-base.sh
fatz: virtual is the minimalistic kernel. This lead to issues compiling the nvidia kernel drivers

build/setup.d/20-install-packages.sh
fatz: generic should have all of them. If someone depends on something special it would be great to have a test for this.

However afaik it was added while i'm already worked on the branch. Therefor i will give the virtual kernel + extra-virtual another try. Maybe this combination is enough for the nvidia driver

build/setup.d/10-upgrade-base.sh
fatz: forget what i sad.... my bad. 

Working on a fixed version

build/setup.d/20-install-packages.sh
fatz: indeed not needed. Tests just got through.

fatz: Ok i'd started the tests on a p2 instance which leads to the use of nvidia-docker. In this case i ran into the symlink issue `libapparmor.so.1 -> libapparmor.so.1.1.0`. With docker 1.12... installed this problem does not exist.

It cost me a lot of time finding this now. Question is what is bad about using a newer minor version even if its already on patch 6

szuecs: Docker is a beast and has to be proven working in all cases for us. We hit with every version change a lot of problems with docker, so if there is no reason then I would prefer no upgrade. I am ok, if there is a reason to upgrade that you upgrade it and it seems that you have a valid reason.

fatz: another solution would be to let the Docker.py runtime check for links in volume sources and resolve them. But TBH that doesn't seem to be a better solution.

fatz:
because i struggled at exactly this point. Im currently improve the test.sh to be able to test different instance types.

Just being in the last tests commit will follow in a few minutes

fatz:
@szuecs anything else? I now added a feature to `test.sh` to test different instance types.

szuecs:
@fatz I have no idea if I can review test.sh. I have to trust it somehow, because the diff looks so ugly.

test.sh
```diff
@@ -5,16 +5,19 @@ set -e
...
 	if [ $DRY_RUN = false ]; then
-		echo "Terminating server..."
-		aws ec2 terminate-instances --region $region --instance-ids $instanceid > /dev/null
-		delete_test_volumes
-		delete_profile_for_volume_attachment
-		if [ -n "$mint_bucket" ]; then
-			taupageyamlfile="$(pwd)/test-userdata.yaml"
-			echo "change mint-bucket to a example again"
-			sed -i "1,$ s/mint_bucket.*$/mint_bucket:\ S3-MINT-BUCKET/" $taupageyamlfile
+		if [ "$?" -gt 0 ] || [ ${1:-false} = true ]; then
```
dehora: @fatz do you need this if check to test the new drivers? If it's not strictly needed, maybe back it out to reduce the diff.

test.sh
```diff
@@ -59,120 +59,134 @@ fi
...
-    --security-group-ids $security_group \
-    --subnet-id $subnet \
-    --output json \
-    --region $region \
-    --user-data file://$(pwd)/test-userdata.yaml)
-
-instanceid=$(echo $result | jq .Instances\[0\].InstanceId | sed 's/"//g')
-echo "Instance: $instanceid"
+if [ -z $testinstance_types ]; then
+	# backward compability
```
dehora: @fatz `compability` -> `compatibility` but you could probably just remove the comment

test.sh
```diff
@@ -59,120 +59,134 @@ fi
...
+	testinstance_types="
+$instance_type
+"
```
dehora: @fatz are the newlines intentional here? just wanted to check you're happy with this (my first thought was `testinstance_types=($instance_type)` but i might not understand the intent)

dehora:
> @fatz I have no idea if I can review test.sh. I have to trust it somehow, because the diff looks so ugly.

@szuecs try https://github.com/zalando-stups/taupage/pull/365/files?w=1, it might help reduce the size. I left a couple of comments.

szuecs:
:+1:

hjacobs:
:+1:

test.sh
elezar: I suppose this is to allow for the looping over `testinstance_types`. @fatz one thing that I noticed is that the test is then not run using the c3 instances (which seemed to be the default behaviour before)?

hjacobs:
Let's resolve the "nit-picks" from @dehora and it's ready to merge :smile:

elezar:
@fatz thanks for the work here! I agree with @hjacobs. Let's at least get the typos addressed, and then we can merged.

test.sh
fatz: @dehora well we can leave out it. But i noticed that the behaviour might be different depending on the instance type. Therefore i added the possibility to check different instance types

fatz: but however using "$?" does not lead to the expected result... fixed it now

test.sh
fatz: dunno what brought me to this idea. testinstance_types="$instance_type" is fairly enough

> one thing that I noticed is that the test is then not run using the c3 instances
i added the c3 to the config-stups-exampl.sh
we can also always add it to the list but IMO we than should check for duplicates. So I didn't want to overengineer this.

config-stups-example.sh
```diff
@@ -17,6 +17,7 @@ instance_type="c3.large"
 testinstance_types="
 t2.micro
 m4.xlarge
+c3.large
```
elezar: what about:
```bash
testinstance_types="
$instance_type
t2.micro
m4.xlarge
p2.xlarge
```

fatz: 🤦‍♂️ thx @elezar

hjacobs:
Ready to merge?

hjacobs:
:+1:

elezar:
:+1:

hjacobs:
@fatz thanks for all the work!

retnuh:
Thanks to everyone who contributed to this, this should help a load of teams!!

PR #371, opened 2017-03-02
Install ena kernel module

CyberDem0n:
It's needed to enable Enhanced Networking with the Elastic Network Adapter (ENA) on Linux Instances in a VPC on instance types R4 and m4.16xlarge.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html#enhanced-networking-ena-ubuntu

### DIFF
--- a/build/setup.d/10-upgrade-base.sh
+++ b/build/setup.d/10-upgrade-base.sh
@@ -32,14 +32,22 @@ EOF
 }
 
 pushd /usr/src
-# download and configure ixgbevf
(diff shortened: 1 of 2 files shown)


Raffo:
Can you do a checksum on the downloaded binaries?

a1exsh:
:+1:

Raffo:
👍

PR #372, opened 2017-03-03
bugfix: forgot to rename drv->module when refactoring

CyberDem0n:
dkms.conf: Error! No 'PACKAGE_NAME' directive specified.


Jan-M:
:+1:

PR #374, opened 2017-03-03
bugfix: fix dkms.config for ENA

CyberDem0n:
wrong version was set and source directory was also wrong

It was causing module build to fail silently...


CyberDem0n:
:+1:

PR #383, opened 2017-03-31
use iid as default

tuxlife:
Change defautl behavior of `registry_login` and change to pierone iid login


tuxlife:
@hjacobs Can we remove the hole `get_token` function from `taupage/__init__.py`?

hjacobs:
@tuxlife yes, if it's no longer used in `Docker.py`...

hjacobs:
:+1:

tuxlife:
:+1:

PR #386, opened 2017-04-04
Kernel upgrade: utopic->xenial

CyberDem0n:
AWS recommend to run Ubuntu 16.04 for NVMe support: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html
We don't want to change base image, but it make sense at least upgrade kernel to xenial.


szuecs:
:+1:

CyberDem0n:
:+1:

Issue #392, opened 2017-05-08
Non-Deterministic User Creation


jbspeakr:
System user creation is currently not done in a deterministic way.

E.g. adding support for nvidia in #365 broke taupage backwards-compatibility by introducing a new system user ([through a dependency](https://github.com/zalando-stups/taupage/blob/9d644b8db803da7b4ce40aa58d8fe56827c30d8f/build/setup.d/40-nvidia.sh)) before the actual (oldschool) [application user](https://github.com/zalando-stups/taupage/blob/9d644b8db803da7b4ce40aa58d8fe56827c30d8f/build/setup.d/85-add-application-user.sh) (that is used to run application containers) gets introduced to the system. 

This becomes a problem when e.g. persisting data via EBS volumes, as the persisted data (after updating to the latest taugape AMI) is not accessible for the application user anymore. Reason for this is that the new nvidia-docker user takes over the formerly assigned UID.

System UIDs should be created in a deterministic way. They must not change.
Application User ID should be 999 and the first user created.

fatz:
i think the nvidia script can run after 85. But has to be tested. 

However this might happen in the future again as long as the userid is not fixed to a special number. as @jbspeakr already mentioned.

Which also leads to the point of a missing test for that.

jbspeakr:
would be pleased if anyone could help with testing this.

PR #393, opened 2017-05-08
set application user id explicitly…

jbspeakr:
… and move implicit nvidia-docker user creation to where it does not interfere 
-> see #392


hjacobs:
Maybe add a test to check that the uid is as expected (spec tests are in `tests/` folder).

jbspeakr:
of course, it's on its way 😎

jbspeakr:
are the serverspec tests actually run by travis? couldn't find a corresponding config/ output...

a1exsh:
:+1:

hjacobs:
@jbspeakr the tests are run by our internal Go CD pipeline (it actually creates an AMI and starts up an EC2 instance).

jbspeakr:
@hjacobs ah, that's nice – so I'll lay checking that pipeline into your hands ;D

hjacobs:
:+1:

PR #398, opened 2017-05-12
Fix logrotation for docker and etcd

mikkeloscar:
Logrotation introduced in #397 was not working because of wrong file permission of the logrotate file and an invalid size definition.

This fixes the logrotation also for etcd which also had the wrong file permissions (introduced in #355)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #403, opened 2017-05-30
add copy_from_start to Scalyr Agent config - this will fix #402

### LINKED ISSUE #402: Scalyr agent should follow log files
Currently if a log file rotates the Scalyr agent doesn't pick it up anymore. This leads to corner cases where the Scalyr agent can't read the last log lines in the "current" application.log.

The Scalyr agent should follow the log files to prevent that from happening.

apfeiffer85:
Modify the Scalyr Agent config to read the log files from the beginning and not just only the last / latest log lines. This will fix #402


mohabusama:
👍

apfeiffer85:
👍

PR #405, opened 2017-06-12
Use compliant images in tests

mikkeloscar:
Use images which are build and served in a compliant way. E.g. has correct scm-source.json and originates from our registry.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #406, opened 2017-06-13
Add image sources for images used in tests

mikkeloscar:
This adds the sources of the two docker images which we use in the tests. By having them here rather than in personal repositories we can build them with a valid `scm-source.json`.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #409, opened 2017-06-13
Fix tiny-docker-http-test image

mikkeloscar:
Adds the changes from this PR https://github.com/hjacobs/tiny-docker-http-test/pull/2 which we have been using since https://github.com/zalando-stups/taupage/commit/d80ccb648f0659fbbfb13037c5befce217df64ea anyway.

It uses an ubuntu base image rather than scratch, because mounting `/usr/bin/docker` like we do when running this image, does not work unless you have stuff like libpthread, libdevmapper etc. in the container.


hjacobs:
:+1:

mikkeloscar:
:+1:

PR #410, opened 2017-06-14
Update test image versions

mikkeloscar:
Use latest images in the test.


mikkeloscar:
:+1:

szuecs:
:+1: I would be more happy if we could use semantic versioning for everything and prefix with "v"

szuecs:
:+1:

Issue #411, opened 2017-06-16
be forgiving on erase_on_boot not present and ec2:DescribeTags and ec2:DeleteTags not available


lotharschulz:
reiterating on #388: I can't see why this feature can't be implemented backwards compatible.

https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L229 could be taken out of if scope.

cc @hjacobs as we discussed this morning about that.

Raffo:
Can you please create a PR for this? There was an original idea of what to implement in #388 from @dryewo .

lotharschulz:
@Raffo : I propose to tackle this with https://github.com/zalando-stups/taupage/pull/412

Raffo:
This was implemented in #413

PR #412, opened 2017-06-27
set erase_tag_set to true even if list of tags is empty

lotharschulz:
'erase_tag_set = True' even if list of tags is empty

related: #411


hjacobs:
This change looks dangerous to me (without knowing the details)...

mikkeloscar:
This would always erase the volume if we don't find any tags.

Isn't this the opposite of what you want?

Raffo:
I thought the same of what @mikkeloscar and @hjacobs said, this is also why I suggest doing something like @dryewo originally suggested in the issue.

lotharschulz:
learned today about `already_mounted` variable (https://github.com/lotharschulz/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L252) is most likely different than I expected it.

PR #414, opened 2017-06-28
Switch download url for ixgbevf driver

mikkeloscar:
Intel seem to have moved the location of the driver, so it now lives under a different path.

The md5sum is still the same.

Fixes builds which are currently failing because the old URL returns 404.


mikkeloscar:
:+1:

linki:
👍

Issue #415, opened 2017-07-05
Support AWS X-Ray Daemon


grahamar:
It would be great to have the option of using AWS X-Ray via the [daemon](http://docs.aws.amazon.com/xray/latest/devguide/xray-daemon.html).

PR #416, opened 2017-07-05
Add support for AWS X-Ray daemon

grahamar:
Install and start (depending on config 'xray_enabled') AWS X-Ray daemon
http://docs.aws.amazon.com/xray/latest/devguide/xray-daemon.html

AWS X-Ray provides distributed tracing. It's cheaper for us to use X-Ray then spin up an EC2 instance for something similar like Zipkin.

We'd like to instrument our services to send HTTP tracing data (Host, Method, URL, Client IP, Duration & Status Code) to AWS X-Ray for debugging & alerting purposes.

### DIFF
--- a/build/setup.d/88-install-x-ray-daemon.sh
+++ b/build/setup.d/88-install-x-ray-daemon.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+mkdir /tmp/x-ray-setup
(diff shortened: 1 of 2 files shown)


grahamar:
Fixes #415

tuxlife:
:+1:

Raffo:
👍

Issue #417, opened 2017-07-06
Missing logs on log rotation


cpeik:
Background
---

We see missing logs for a couple of seconds when log rotation is done.

```
cpeikert@ip-172-31-134-88:/var/log$ tail -n 1 application.log.1 | cut -b 1-15
Jul  6 09:17:14
cpeikert@ip-172-31-134-88:/var/log$ head -n 1 application.log | cut -b 1-15
Jul  6 09:17:22
```

I assume this is because of the default `logrotate` settings that we cannot override via Taupage config. [There](https://github.com/zalando-stups/taupage/blob/master/runtime/etc/logrotate.d/application) `copytruncate` is configured:

>        copytruncate
>               Truncate the original log file to zero size in place after creating a copy, instead of moving the old log file and optionally creating a new one.  It can be used when some program cannot be told to close  its  logfile
>               and  thus  might  continue  writing  (appending) to the previous log file forever.  Note that there is a very small time slice between copying the file and truncating it, so some logging data might be lost.  When this
>               option is used, the create option will have no effect, as the old log file stays in place.

 
Proposed Solution
---

Make `copytruncate` configurable.

BTW:

We would like to unset `delaycompress`, too. It almost doubles the disk usage for logs and thus we can keep less logs because of this. 

>        delaycompress
>               Postpone  compression  of  the previous log file to the next rotation cycle.  This only has effect when used in combination with compress.  It can be used when some program cannot be told to close its logfile and thus
>               might continue writing to the previous log file for some time.

cpeik:
https://github.com/zalando-stups/taupage/issues/254 is related

tuxlife:
merged

PR #419, opened 2017-07-12
make copytruncate and delaycompress behavior configurable

tuxlife:
Fix for #417

### DIFF
--- a/runtime/opt/taupage/init.d/86-adjust-application-logrotate
+++ b/runtime/opt/taupage/init.d/86-adjust-application-logrotate
@@ -30,6 +30,18 @@ if [ -n "$config_application_logrotate_rotate" ]; then
         sed -i "s/\brotate\b.*$/rotate $config_application_logrotate_rotate/g" $application_logrotate
 fi
 
+#disable the copytruncate and use the default behavior for logrotate
+if [ -n "$config_application_logrotate_disable_copytruncate" = True ]; then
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/init.d/86-adjust-application-logrotate
```diff
@@ -30,6 +30,18 @@ if [ -n "$config_application_logrotate_rotate" ]; then
         sed -i "s/\brotate\b.*$/rotate $config_application_logrotate_rotate/g" $application_logrotate
 fi
 
+#disable the copytruncate and use the default behavior for logrotate
+if [ -n "$config_application_logrotate_disable_copytruncate" = True ]; then
```
Raffo: You don't need the `= True`, right?

Raffo:
Only one comment, the rest look fine.

runtime/opt/taupage/init.d/86-adjust-application-logrotate
tuxlife: If I don't use `True` it would be also disabled with `application_logrotate_disable_copytruncate: false`

Raffo: Okay!

Raffo:
👍

tuxlife:
:+1:

PR #421, opened 2017-07-12
Custom routing to predefined NAT when run in DMZ without public ip

CyberDem0n:
We have a use-case when we want to run instances in DMZ or public subnet but without public ips. Elastic ip would be assigned later to one of the instances. Without public ip instance will not be able to initialize (download docker, push logs, use AWS api, etc...). To solve this problem we need to create a separate routing table for outgoing https traffic. Such traffic would be routed to a custom nat gateway.
Mappings between subnets and nat gateways would come from UserData and populated by senza.


tuxlife:
:+1:

Raffo:
👍

PR #424, opened 2017-07-14
Wait for devices before checking that raid exists

CyberDem0n:
If devices were already part of existing raid, kernel could build the raid automatically. In this case we can't create the new one with mdadm.

Also use the right name when performing check that device is already attached.


tuxlife:
:+1:

Raffo:
👍

PR #426, opened 2017-07-18
Format partition with nodiscard

CyberDem0n:
Discard is useful only for SSD volumes, but all instance volumes already discarded by AWS, so it doesn't make sense to discard it second time.


Jan-M:
:+1:

alexeyklyukin:
👍

Issue #425, opened 2017-07-18
The daily logrotate starves the system resources


mo-gr:
A while ago, the hourly log rotate was re-niced. But the daily log rotate is still running at high priority.

This is the output from a recent taupage system:

```sh
$ diff /etc/cron.daily/logrotate /etc/cron.hourly/logrotate 
14c14
< /usr/sbin/logrotate /etc/logrotate.conf
---
> nice -n 19 ionice -c3 /usr/sbin/logrotate /etc/logrotate.conf
```

This also opens the question, why the daily and the hourly logrotate to exactly the same, just at different nice-levels. Maybe it's worth considering to delete the daily log rotate?

szuecs:
@mo-gr please review the PR

vigneshshanmugam:
Had a discussion offline with @szuecs and tested with the latest taupage on tailor with this fix. Still seeing the latency spikes during the same time as before. 
<img width="430" alt="screen shot 2017-07-28 at 10 54 15" src="https://user-images.githubusercontent.com/3902525/28710111-239ffd82-7383-11e7-94bf-6cb2246a60c8.png">

szuecs:
These spikes are ~every 10 minutes, for sure not caused by logrotate daily. Can you create a new issue (and close this one) or change the issue accordingly such that the subject reflects the topic?
Thanks!

vigneshshanmugam:
changed the graph timeline to show only the impact caused by the rotation. Check the full graph for 6 hours

<img width="521" alt="screen shot 2017-07-28 at 11 46 02" src="https://user-images.githubusercontent.com/3902525/28712153-61aa8442-738a-11e7-981c-4e25e98bf3dd.png">

mo-gr:
I agree, that the daily log rotate is probably not the cause. So I'm happy to close this. But I still believe something in the daily cron job is causing this.

Raffo:
Maybe related: https://github.com/zalando-stups/taupage/pull/432

PR #427, opened 2017-07-18
use nice to lower the prio for daily logrotate

szuecs:
as suggested in #425 we should use nice to lower the priority


tuxlife:
:+1:

mo-gr:
:+1:

mo-gr:
However, also think about whether it is really necessary to have daily *and* hourly log rotation. One should be enough

szuecs:
@mo-gr it's the default

PR #428, opened 2017-07-19
Call iptables with -w argument

CyberDem0n:
To prevent multiple instances of the program from running concurrently, an attempt will be made to obtain an exclusive lock at launch. By default, the program will exit if the lock cannot be obtained.

Without `-w` iptables could fail with:
```
"Another app is currently holding the xtables lock. Perhaps you want to use the -w option?"
```


tuxlife:
:+1:

mikkeloscar:
:+1:

PR #435, opened 2017-08-11
Nodiscard and 3 attempts to find volumes

CyberDem0n:
* Bugfix: extended-options formatting options were not applied when container was starting as a root user.
* Do 3 attempts with 60 seconds interval to find "available" volumes


tuxlife:
:+1:

CyberDem0n:
:+1:

PR #438, opened 2017-08-17
error message adaption

lotharschulz:
Scripts in init.d folder are started and should return the adapted error message in case the fail at some point.


hjacobs:
@lotharschulz what is the intent for this change? I don't really see the added value of the changed message :thinking:

lotharschulz:
Well, the error message is misleading IMO. As of now the message intents the script in question did not start. I ran into cases when the script started and _did not finish successfully_. 
This caused the error message in syslog. I believe this is true for most maybe all ERROR cases.

tuxlife:
:+1:

lotharschulz:
👍

PR #447, opened 2017-09-12
Send ntp traffic via NAT-GW

CyberDem0n:
if node is running in public subnet without public ip it can't sync time because ntp servers are not reachable.
For ntp traffic we will do the same trick as for https, i.e. send it via NAT-GW


tuxlife:
:+1:

CyberDem0n:
:+1:

PR #448, opened 2017-09-12
Change cron start

### LINKED ISSUE #440: Change cron start time
This relates to #439 

I think it makes sense to change the cron-tab such that daily, weekly and monthly cron jobs run in the middle of the night instead of during the morning.

I would suggest moving all cron jobs from 6 UCT to 1 UCT

mo-gr:
This adds the crontab from a running system and changes the startup time from daily/weekly/monthly from 6:XX UTC to 1:XX UTC.

This should fix #440

### DIFF
--- a/runtime/etc/crontab
+++ b/runtime/etc/crontab
@@ -1,0 +1,15 @@
+# /etc/crontab: system-wide crontab
+# Unlike any other crontab you don't have to run the `crontab'
+# command to install the new version when you edit this file
+# and files in /etc/cron.d. These files also have username fields,
+# that none of the other crontabs do.
+
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+# m h dom mon dow user	command
+17 *	* * *	root    cd / && run-parts --report /etc/cron.hourly
+25 6	* * *	root	test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily )
+47 6	* * 7	root	test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.weekly )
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

szuecs:
I guess #449 is fixing your issue, not sure if yours will change anything you described earlier.

Joneser:
This would be nice too I think, it means spikes (if any) caused by the scheduled cron should happen in the middle of the night, thereby reducing customer impact

Joneser:
@szuecs Is there anything needed from teams using taupage to update to the latest AMI aside from just deploying a new stack? We created a new stack after the last release which I believe contained #449 and we still experienced the same CPU spikes at 06:30 UTC.  

If there is nothing else to do, maybe we can still merge this as it looks like the previous change didn't resolve the problem.

Joneser:
:+1:

Joneser:
@szuecs is there a regular time that new versions of taupage are published? It looks like there hasn't been a new release since this was merged

szuecs:
@Joneser We don't have a real release cycle, because we work only on this if there is a need stated by a user. Normally we first share the automated build internally with a user that can confirm, the change works for her. Later we publish it. I normally like to publish these things Tuesday morning. I will have a look if I can publish it, now.

szuecs:
FYI: I shared it internal with your team

shuhei:
It turned out that ami-9706b1f8 does not rotate logs at all. Hourly cron job does not seem to be working at all. It was good that the internal build was shared with only our team.

szuecs:
I unshared it

mikkeloscar:
@szuecs should we also revert the changes to ensure the don't go out in the next release?

aermakov-zalando:
@shuhei could you please provide syslog content from an instance? I've just tested it and I don't see any issues with the hourly cron jobs.

aermakov-zalando:
Looks like you were right. No idea why it worked for me on a test instance, maybe while I was editing the crontab my editor helpfully fixed the permissions?

Issue #439, opened 2017-08-23
Daily cron job `apt` starves our application from resources


mo-gr:
taupage contains the default ubuntu cron jobs. One of them is `apt` (`/etc/cron.d/apt`). This daily job is very resource hungry and even though it es internally reniced (to 10), it creates so much load on the system, that we observe regular instance failures due to missed health checks on our application (tailor).

Every morning between 8:25 and 8:55 we observe huge latency spikes, which frequently lead to ASGs terminating instances.

The aforementioned apt-job is executed at 8:25 and does a random sleep between 0 and 30 minutes.

When we manually disable the apt-job, we neither see the huge latency spikes nor do we see the ASG terminating instances.

Since taupage is intended as an "immutable" base OS, I don't see any reason to run the daily `apt` job in the first place. Therefore I would recommend completely disabling this job on taupage. 

I would also appreciate if you could check all the other daily, weekly and monthly cron jobs and disable those, that are not necessary due to taupages nature.

Joneser:
Is there any plan for this to be worked on? We are seeing the same issues mentioned above and would love to see it resolved, the solution mentioned in #440 would certainly mitigate our problems.

aermakov-zalando:
Fixed in #449.

PR #449, opened 2017-09-12
Disable periodic apt updates

### LINKED ISSUE #434: Daily cronjobs causing huge latency spikes
We initially suspected log rotation in #425 But that does not seem to be the cause.

When analysing this, keep in mind, that our application acts as a proxy for many other applications, that are also deployed on taupage.

So whatever we test for our application, will very likely not fix this, since during the daily cron-timeslot, all the applications, that we are proxying to will also be affected.

aermakov-zalando:
As reported previously (#434), this can consume a lot of resources on small instances and cause the ASG to terminate instances. We don't need this, so let's just disable the corresponding cronjobs.

Additionally, stuff like `popularity-contest` should probably not run on our production instances, so let's remove that as well.

Fix #434, #439.

### DIFF
--- a/build/setup.d/11-remove-unneeded-packages.sh
+++ b/build/setup.d/11-remove-unneeded-packages.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -euo pipefail
+
+pkgs="
+popularity-contest
+update-notifier-common
+"
+
+echo "Removing packages..."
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

aermakov-zalando:
👍

PR #450, opened 2017-09-12
Remove shell options from remove-unattended-packages.sh

aermakov-zalando:
The `setup.d` scripts are just sourced with `.`, not started normally. This means we shouldn't change options, otherwise we might break something else.


aermakov-zalando:
👍

szuecs:
:+1:

PR #451, opened 2017-09-13
Updating Docker version

danmanirl:
Update the docker version to enable some of the newer features. (i.e. support for docker-compose version 3)

### DIFF
--- a/build/setup.d/20-install-packages.sh
+++ b/build/setup.d/20-install-packages.sh
@@ -6,7 +6,7 @@
 pkgs="
 auditd
 build-essential
-docker-engine=1.12.6-0~ubuntu-trusty
+docker-engine=17.05.0~ce-0~ubuntu-trusty
 gcc
 htop
 iproute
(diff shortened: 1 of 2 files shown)


hjacobs:
I would be hesitant to just bump the Docker version. We had problems with Docker versions in the past and I would avoid upgrading a known-good version for some "unofficial" use case like Docker Compose. What is your use case of Docker Compose? Running it as build/test on CI (Jenkins)?

danmanirl:
@hjacobs Thanks for getting back so quickly! 

So we are using docker-compose to bring up kafka and zookeeper for integration tests within "cloud-kraken" instances in jenkins.  We wanted to use newer features in docker-compose like networking and default variable substitution which I thought was only possible from the docker-compose file version 3.  

As it turns out, these features are also available in version "2.1" which only requires docker version >=12.6 so we don't need an upgraded docker after all.  Obviously there are some nice things in file version 3 (like support for `docker stack`) which would be nice to have in the future but it is not something we require.

Closing the PR as it is no longer required.

hjacobs:
@danmanirl please also check out our internal Continuous Delivery Platform (CDP) which runs builds in an isolated VM with a recent Docker version, example: https://github.com/zalando-stups/pierone/blob/master/delivery.yaml

PR #452, opened 2017-09-25
Revert "Change cron start (#448)"

aermakov-zalando:
The [comment](https://github.com/zalando-stups/taupage/pull/448#issuecomment-330992842) was right. I'm not sure why it worked in my testing, but cron refuses to read `/etc/crontab` with its default permissions of 664. Instead of tweaking permissions post-install, let's just revert the change.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #454, opened 2017-09-27
Fix download links for the NIC drivers

aermakov-zalando:
Intel changed the URL, update the download ID. Additionally, use curl --fail so it's clear why it failed.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #457, opened 2017-09-27
remove whitespaces and delete the old install instana agent script

apfeiffer85:
remove whitespaces and delete the old install instana agent script. 
In the last PR it was using the wrong(old) script so therefore we have  to delete it again. :)


vetinari:
:+1:

apfeiffer85:
👍

PR #464, opened 2017-10-09
Disable instana if Appd is enabled

rajatparida86:
Don't merge to master as is used for POC.

Disable instana if Appd is enabled


rajatparida86:
Don't merge to master

hjacobs:
I guess we can close this as the PoC is over.

PR #468, opened 2017-10-11
Instana cassandra stupstest

rajatparida86:
Added instana host, port and key to cfg file at run time


hjacobs:
I guess we can close this as the PoC is over.

PR #470, opened 2017-10-23
Fix permissions on /etc/cron.d/logrotate

aermakov-zalando:
File permissions were broken in the previous change, restore to sane values.


aermakov-zalando:
👍

tuxlife:
:+1:

PR #472, opened 2017-11-01
Update nvidia-driver to support Tesla Volta cards

rbrthogan:
Version bump of nvidia-driver to 384.81 support Tesla V100 cards used in AWS p3 instances

### DIFF
--- a/build/setup.d/95-nvidia.sh
+++ b/build/setup.d/95-nvidia.sh
@@ -3,7 +3,7 @@
 # See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
 # But we're using the run script to not depend on build tools
 DRIVER_ARCH="Linux-x86_64"
-DRIVER_VERSION="375.26"
+DRIVER_VERSION="384.81"
 DRIVER_FILENAME="NVIDIA-${DRIVER_ARCH}-${DRIVER_VERSION}.run"
 DRIVER_CHECKSUM="d60819b2e377398c7296999ab5e7c1a4"
 
--- a/build/setup.d/95-nvidia.sh
+++ b/build/setup.d/95-nvidia.sh
@@ -3,9 +3,9 @@
 # See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
 # But we're using the run script to not depend on build tools
 DRIVER_ARCH="Linux-x86_64"
-DRIVER_VERSION="384.81"
+DRIVER_VERSION="384.90"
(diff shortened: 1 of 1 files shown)


build/setup.d/95-nvidia.sh
```diff
@@ -3,7 +3,7 @@
 # See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
 # But we're using the run script to not depend on build tools
 DRIVER_ARCH="Linux-x86_64"
-DRIVER_VERSION="375.26"
+DRIVER_VERSION="384.81"
 DRIVER_FILENAME="NVIDIA-${DRIVER_ARCH}-${DRIVER_VERSION}.run"
 DRIVER_CHECKSUM="d60819b2e377398c7296999ab5e7c1a4"
```
aermakov-zalando: The hash should be updated as well.

build/setup.d/95-nvidia.sh
```diff
@@ -3,7 +3,7 @@
 # See https://github.com/NVIDIA/nvidia-docker/wiki/Deploy-on-Amazon-EC2
 # But we're using the run script to not depend on build tools
 DRIVER_ARCH="Linux-x86_64"
-DRIVER_VERSION="375.26"
+DRIVER_VERSION="384.81"
```
aermakov-zalando: http://us.download.nvidia.com/XFree86/Linux-x86_64/384.81/NVIDIA-Linux-x86_64-384.81.run leads to a 404, are you sure the version number is correct?

rbrthogan: hmmm, ya you're right. I just found the version number here http://www.nvidia.com/download/driverResults.aspx/124728/en-us the download link appears to be different

rbrthogan: it appears from http://www.nvidia.com/object/unix.html that 384.90 is the latest version that supports to same cards as 375.26.

Oddly, the product page seems to say that neither supports Tesla cards, despite having used 375.26 with tesla K80s.

*TLDR* I think it should be 384.90, I will update

build/setup.d/95-nvidia.sh
rbrthogan: will update

chinoxyz:
I am wondering what is the status of this. Is this going to be merged?

Raffo:
Can you fix the build as well?

aermakov-zalando:
I've fixed the build [here](https://github.com/zalando-stups/taupage/pull/479), sorry for the delay.

aermakov-zalando:
Reopening to trigger Travis.

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #477, opened 2017-12-12
Add support for attaching network interfaces based of a tag

imduffy15:
This feature enables you to associate a Static IP to an instance.

Given a network interface exists within the same AZ as your instance,
and the network interface has a name tag, the value of the name tag
can be specified in the taupage config and taupage will then attach
and configure it on boot up.

```
network_interfaces:
    - Name_tag_value
```

This is helpful for applications like ZooKeeper that do not play
nicely behind a load balancer: https://issues.apache.org/jira/browse/ZOOKEEPER-1846

Signed-off-by: Ian Duffy <ian.duffy@zalando.ie>

### DIFF
--- a/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
+++ b/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
@@ -30,25 +30,6 @@ def retry(func):
     return wrapped
 
 
-class CmdException(Exception):
-    def __init__(self, returncode, errmsg):
-        self._returncode = returncode
-        message = "Command returned non-zero exit status {}:\n{}".format(returncode, errmsg)
-        super(CmdException, self).__init__(message)
-
-    @property
-    def returncode(self):
-        return self._returncode
-
-
-def call_command(call, allowed_error_codes=[0]):
-    proc = subprocess.Popen(call, stderr=subprocess.PIPE)
-    stdout, stderr = proc.communicate()
-    if proc.returncode not in allowed_error_codes:
-        raise CmdException(proc.returncode, stderr.decode('utf-8'))
-    return stdout
-
-
 def ec2_client(region):
     return boto.ec2.connect_to_region(region)
 
@@ -84,7 +65,8 @@ def find_network_interface(ec2, name):
                 sys.exit(2)
 
     if len(network_interfaces) > 1:
-        logging.warning('More than one network interface with name %s found.', name)
+        logging.warning(
+            'More than one network interface with name %s found.', name)
         network_interfaces.sort(key=lambda v: v.id)
     return network_interfaces[0].id
 
@@ -123,7 +105,8 @@ def handle_network_interfaces(region, config):
     for index, name in enumerate(network_interfaces):
         try:
             device_index = index + 1
-            attach_network_interface(ec2, find_network_interface(ec2, name), device_index)
+            attach_network_interface(
+                ec2, find_network_interface(ec2, name), device_index)
             wait_for_network_interface_attachment(device_index)
         except Exception as e:
             logging.exception(e)
@@ -203,26 +186,28 @@ def main():
         # Note, we do not run dhclient on eth0 as this may affect network connectivity
         # of the instance
         for network_interface in network_interfaces[1:]:
-            call_command(["dhclient", str(network_interface)])
+            subprocess.check_call(["dhclient", str(network_interface)])
         route_tables = []
 
         # Here we implement source-based routing, according to the serverfault post linked above
         for device_index in range(0, len(config.get("network_interfaces")) + 1):
-            route_tables.append("{} eth{}".format(device_index + 1, device_index))
+            route_tables.append("{} eth{}".format(
+                device_index + 1, device_index))
 
         with open("/etc/iproute2/rt_tables", "w") as rt_tables:
             rt_tables.write("\n".join(route_tables))
 
         for network_interface in network_interfaces:
-            interface = netifaces.ifaddresses(network_interface)[netifaces.AF_INET][0]
+            interface = netifaces.ifaddresses(network_interface)[
+                netifaces.AF_INET][0]
             ip = interface['addr']
             subnet_cidr = str(IPAddress(interface["netmask"]).netmask_bits())
-            call_command(["ip", "route", "add", "default", "via", default_gateway, "dev",
-                          network_interface, "table", network_interface], allowed_error_codes=[0, 2])
-            call_command(["ip", "route", "add", subnet_cidr, "dev", network_interface,
-                          "src", ip, "table", network_interface], allowed_error_codes=[0, 2])
-            call_command(["ip", "rule", "add", "from", ip, "table",
-                          network_interface], allowed_error_codes=[0, 2])
+            subprocess.check_call(["ip", "route", "add", "default", "via", default_gateway, "dev",
+                                   network_interface, "table", network_interface])
+            subprocess.check_call(["ip", "route", "add", subnet_cidr, "dev", network_interface,
+                                   "src", ip, "table", network_interface])
+            subprocess.check_call(["ip", "rule", "add", "from", ip, "table",
+                                   network_interface])
 
 
 if __name__ == '__main__':
--- a/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
+++ b/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
@@ -11,22 +11,7 @@ import subprocess
 import netifaces
 from netaddr import IPAddress
 
-from taupage import configure_logging, get_config
-
-
-def instance_id():
-    """Helper to return theid for the current instance"""
-    return boto.utils.get_instance_metadata()['instance-id']
-
-
-def detect_region():
-    """Helper to return the region for the current instance"""
-    return boto.utils.get_instance_metadata()['placement']['availability-zone'][:-1]
-
-
-def zone():
-    """Helper to return the AZ for the current instance"""
-    return boto.utils.get_instance_metadata()['placement']['availability-zone']
+from taupage import configure_logging, get_config, get_instance_id, get_region, get_availability_zone
 
 
 def retry(func):
@@ -81,7 +66,7 @@ def find_network_interface(ec2, name):
             network_interfaces = list(get_all_network_interfaces(ec2, {
                 'tag:Name': name,
                 'status': 'available',
-                'availability-zone': zone()
+                'availability-zone': get_availability_zone()
             }))
         except Exception as e:
             logging.exception(e)
@@ -128,7 +113,7 @@ def attach_network_interface(ec2, network_interface_id, device_index):
     ec2.attach_network_interface(
         device_index=device_index,
         network_interface_id=network_interface_id,
-        instance_id=instance_id()
+        instance_id=get_instance_id()
     )
 
 
@@ -185,7 +170,7 @@ def main():
     else:
         configure_logging(logging.INFO)
 
-    current_region = args.region if args.region else detect_region()
+    current_region = args.region if args.region else get_region()
 
     # Load configuration from YAML file
     config = get_config(args.filename)
@@ -218,7 +203,7 @@ def main():
         # Note, we do not run dhclient on eth0 as this may affect network connectivity
         # of the instance
         for network_interface in network_interfaces[1:]:
-            call_command(["dhclient", str(network_interface)], allowed_error_codes=[0, 2])
+            call_command(["dhclient", str(network_interface)])
         route_tables = []
 
         # Here we implement source-based routing, according to the serverfault post linked above
--- a/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
+++ b/runtime/opt/taupage/init.d/04-configure-network-interfaces.py
@@ -1,0 +1,244 @@
+#!/usr/bin/env python
+
+import argparse
+import logging
+
+import boto.ec2
+import boto.utils
+import time
+import sys
+import subprocess
+import netifaces
+from netaddr import IPAddress
+
+from taupage import configure_logging, get_config
+
+
+def instance_id():
+    """Helper to return theid for the current instance"""
+    return boto.utils.get_instance_metadata()['instance-id']
+
+
+def detect_region():
+    """Helper to return the region for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone'][:-1]
+
+
+def zone():
+    """Helper to return the AZ for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone']
+
+
+def retry(func):
+    def wrapped(*args, **kwargs):
+        count = 0
+        while True:
+            try:
+                return func(*args, **kwargs)
+            except boto.exception.BotoServerError as e:
+                if count >= 10 or str(e.error_code) not in ('Throttling', 'RequestLimitExceeded'):
+                    raise
+                logging.info('Throttling AWS API requests...')
+                time.sleep(2 ** count * 0.5)
+                count += 1
+
+    return wrapped
+
+
+class CmdException(Exception):
+    def __init__(self, returncode, errmsg):
+        self._returncode = returncode
+        message = "Command returned non-zero exit status {}:\n{}".format(returncode, errmsg)
+        super(CmdException, self).__init__(message)
+
+    @property
+    def returncode(self):
+        return self._returncode
+
+
+def call_command(call, allowed_error_codes=[0]):
+    proc = subprocess.Popen(call, stderr=subprocess.PIPE)
+    stdout, stderr = proc.communicate()
+    if proc.returncode not in allowed_error_codes:
+        raise CmdException(proc.returncode, stderr.decode('utf-8'))
+    return stdout
+
+
+def ec2_client(region):
+    return boto.ec2.connect_to_region(region)
+
+
+@retry
+def get_all_network_interfaces(ec2, filters):
+    return ec2.get_all_network_interfaces(filters=filters)
+
+
+def find_network_interface(ec2, name):
+    tries = 10
+    network_interfaces = []
+    while not network_interfaces:
+        try:
+            network_interfaces = list(get_all_network_interfaces(ec2, {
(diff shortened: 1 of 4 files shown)


runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+        '--region',
+        dest='region',
+        help='uses a specific AWS region instead of querying the instance metadata'
+    )
+
+    return parser.parse_args()
+
+
+def main():
+    args = process_arguments()
```
mfelsche: the taupage init script is called without any arguments:
https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.sh#L24

So have this argument parsing here might be overkill.

If you want to have this script available for calling it manually (with different arguments/options), maybe place it in `runtime/opt/taupage/bin/` and make the one in `/opt/taupage/init.d/` a bash script just calling the one in bin.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+import time
+import sys
+import subprocess
+import netifaces
+from netaddr import IPAddress
+
+from taupage import configure_logging, get_config
+
+
+def instance_id():
```
mfelsche: the returned values won't change and thus can be cached statically to avoid repeated lookups / AWS API calls.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+    """Helper to return theid for the current instance"""
+    return boto.utils.get_instance_metadata()['instance-id']
+
+
+def detect_region():
```
mfelsche: same here

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+    """Helper to return the region for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone'][:-1]
+
+
+def zone():
+    """Helper to return the AZ for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone']
```
mfelsche: same here

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+            else:
+                sys.exit(2)
+
+    if len(network_interfaces) > 1:
+        logging.warning('More than one network interface with name %s found.', name)
+        network_interfaces.sort(key=lambda v: v.id)
+    return network_interfaces[0].id
+
+
+def wait_for_network_interface_attachment(ec2, device_index):
```
mfelsche: ec2 is not used.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+    """Helper to return the region for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone'][:-1]
+
+
+def zone():
+    """Helper to return the AZ for the current instance"""
+    return boto.utils.get_instance_metadata()['placement']['availability-zone']
+
+
+def retry(func):
```
mfelsche: boto already does AWS API request retrying.
As a suggestion: for some more explicit exception retry, you could use https://pypi.python.org/pypi/backoff/1.4.3

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+            tries -= 1
+            if tries > 0:
+                logging.error(
+                    'Sleeping for 10 seconds and hope a network interface will become "available"')
+                time.sleep(10)
+            else:
+                sys.exit(2)
+
+    if len(network_interfaces) > 1:
+        logging.warning('More than one network interface with name %s found.', name)
```
mfelsche: this doesn't need to be a warning imho

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+        return "eth{}".format(device_index) in netifaces.interfaces()
+
+    attachment = get_attachment()
+    while not attachment:
+        attachment = get_attachment()
+        if not attachment:
+            tries -= 1
+            if tries > 0:
+                logging.error(
+                    'Sleeping for 60 seconds waiting for network interface to attach')
```
mfelsche: `... to be attached`

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def handle_network_interfaces(region, config):
+    ec2 = ec2_client(region)
+    network_interfaces = config.get("network_interfaces", [])
+    for index, name in enumerate(network_interfaces):
+        try:
+            device_index = index + 1
+            attach_network_interface(ec2, find_network_interface(ec2, name), device_index)
+            wait_for_network_interface_attachment(ec2, device_index)
```
mfelsche: If there are > 1 ec2 instances starting up at the same time (normal behaviour for most ASGs we use) this code runs simultaneously and thus two instances might try to attach the same network interface that they found simultaneously.
One of them will thus fail calling `attach_network_interface`, resulting in the taupage instance not finishing its init process properly: https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.sh#L27

I am not sure if this i what you want.

In case `attach_network_interface` fails or `wait_for_network_interface_attachment` fails it might make sense to re-execute: `find_network_interface` to get an available network interface that is not already bound by another ec2 instance, `attach_network_interface` and `wait_for_network_interface_attachment`. (maybe only 3 times)

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+                'status': 'available',
+                'availability-zone': zone()
+            }))
+        except Exception as e:
+            logging.exception(e)
+            sys.exit(2)
+
+        if not network_interfaces:
+            logging.error(
+                'No matching "available" network interfaces volume with name %s found.', name)
```
mfelsche: remove `volume`

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+        with open("/proc/sys/net/ipv4/conf/all/arp_filter", "w") as all_arp_filter:
+            all_arp_filter.write("1")
+        network_interfaces = []
+        default_gateway = netifaces.gateways()['default'][netifaces.AF_INET][0]
+
+        for device_index in range(0, len(config.get("network_interfaces")) + 1):
+            network_interfaces.append("eth{}".format(device_index))
+
+        for network_interface in network_interfaces[1:]:
+            call_command(["dhclient", "{}".format(network_interface)], allowed_error_codes=[0, 2])
```
mfelsche: unnecessary format, `str(...)` should be fine.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,226 @@
...
+        for device_index in range(0, len(config.get("network_interfaces")) + 1):
+            route_tables.append("{} eth{}".format(device_index + 1, device_index))
+
+        with open("/etc/iproute2/rt_tables", "w") as rt_tables:
+            rt_tables.write("\n".join(route_tables))
+
+        for network_interface in network_interfaces:
+            interface = netifaces.ifaddresses(network_interface)[netifaces.AF_INET][0]
+            ip = interface['addr']
+            subnet_cidr = "{}".format(IPAddress(interface["netmask"]).netmask_bits())
```
mfelsche: unnecessary format, `str(...)` should be fine.

mfelsche:
Some comments on why enabling arp_filter, running dhclient and fixing ip routes is necessary would also be nice for maintainability.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: Copied from https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L18 but will adjust

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: Again copied from something we already do in taupage: https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L33

mfelsche: If so, it is fine for me.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
mfelsche: Then it is fine.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: > If there are > 1 ec2 instances starting up at the same time (normal behaviour for most ASGs we use)

I was working under the assumption that it would only be a single instance in the ASG.

Much of the logic here is copied from the mount volume by tag code.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: Again comes what can be seen in the volume script: https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L418

mfelsche: Ok, ignore my comment then. :)

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
mfelsche: If this is fine with other reviewers, then this assumption should be made transparent in the script and docs.

I have to admit that if we could make this script safe for having multiple instances started at the same time going for a group of network_interfaces, with a rather simple change like i suggested above, we should go for it.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: I would consider it a warning, could result in the instance getting an unexpected IP.

Should two interfaces exist within the same AZ, they have the same name and are both available. The instance will attempt to attach the head.

imduffy15:
Thanks for the review @mfelsche I think i've addressed everything or we've had a brief discussion, if there's anything else let me know :)

mikkeloscar:
@imduffy15 can you rebase on master?

tuxlife:
@imduffy15 Can you please splitt this pr in one formatting and one feature pull request.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,244 @@
...
+import sys
+import subprocess
+import netifaces
+from netaddr import IPAddress
+
+from taupage import configure_logging, get_config
+
+
+def instance_id():
+    """Helper to return theid for the current instance"""
```
aermakov-zalando: It might be a better idea to move the common code to the python package in `/usr/local/lib/python3.4/dist-packages/taupage` instead of copying it every single time.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,244 @@
...
+        default_gateway = netifaces.gateways()['default'][netifaces.AF_INET][0]
+
+        for device_index in range(0, len(config.get("network_interfaces")) + 1):
+            network_interfaces.append("eth{}".format(device_index))
+
+        # Run dhclient on all newly created interfaces to enable them to get IPs
+        # Note, we do not run dhclient on eth0 as this may affect network connectivity
+        # of the instance
+        for network_interface in network_interfaces[1:]:
+            call_command(["dhclient", str(network_interface)], allowed_error_codes=[0, 2])
```
aermakov-zalando: Why is exit status 2 fine? Is it documented anywhere?

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
```diff
@@ -0,0 +1,244 @@
...
+                if count >= 10 or str(e.error_code) not in ('Throttling', 'RequestLimitExceeded'):
+                    raise
+                logging.info('Throttling AWS API requests...')
+                time.sleep(2 ** count * 0.5)
+                count += 1
+
+    return wrapped
+
+
+class CmdException(Exception):
```
aermakov-zalando: Do we really need this? There's already `subprocess.check_call` and the corresponding `subprocess.CalledProcessError`.

imduffy15:
Slowly getting around to everyone's comments, sorry for the delay just back from winter holidays today. 

> @imduffy15 can you rebase on master?

@mikkeloscar done :)

> Can you please splitt this pr in one formatting and one feature pull request.

@tuxlife those commits have been dropped - issues appear to have been fixed in master now :)

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: You've spotted a bug here :) this was copied over from the route add command, where an exit code of 2 is OK as it means the route already exists.
PR updated.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: PR updated, thanks for this. Don't know too much about taupage and was trying to limit the amount of things I was touching with this change.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: Again, this comes from `10-prepare-disks.py` https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.d/10-prepare-disks.py#L121

Open to suggestions if there's a better way of handling it.

aermakov-zalando: I'm not sure why it was done like this there, maybe to capture stderr? Since you're not really using it here I'd go with built-in Python classes.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
aermakov-zalando: Could you link the relevant documentation? I couldn't find the exit statuses of `ip route`/`ip rule` documented anywhere at a first glance.

Besides, does it really make sense to just blindly proceed with the next commands instead of failing? The script will only run once during the boot process, so I'd expect it to either configure the instance without any issues or fail and log an error (with the instance being terminated and restarted automatically). The current approach seems more brittle to me — some commands will work, some will silently break but then we'll run the next commands anyway, leaving the instance in an indeterminate state. Are you sure this is absolutely required?

imduffy15: > Could you link the relevant documentation? I couldn't find the exit statuses of ip route/ip rule documented anywhere at a first glance.

I copied this over from @mfelsche work, he may be able to comment on it, I too was unable to find docs relating too it. I did test it out manually though.

> Besides, does it really make sense to just blindly proceed with the next commands instead of failing? The script will only run once during the boot process, so I'd expect it to either configure the instance without any issues or fail and log an error (with the instance being terminated and restarted automatically)

Agreed with this approach, it makes sense to me to adjust to only accept exit code 0.

runtime/opt/taupage/init.d/04-configure-network-interfaces.py
imduffy15: Updated to just use `check_call`. Thanks for the feedback @aermakov-zalando

imduffy15:
👍

regispl:
Is there any chance for this PR to be "revived"? 😉

dehora:
@tuxlife @mikkeloscar @mfelsche  @aermakov-zalando what else needs to happen for this to be merged?

imduffy15:
Hey @regispl 

From my side I will not be doing any follow up work on this PR. The PR was opened months ago, got a review, the feedback was applied and then the reviewal process completely died. I pinged members of team teapot several times, however it never progressed. More than happy for somebody else to take on whatever is required to get it merged in.

Assuming your use case is still Kafka/Zookeeper and your attempting to workaround https://issues.apache.org/jira/browse/ZOOKEEPER-338 the good news is a PR has been opened to fix the ZKClient and its DNS caching ways.... https://github.com/apache/zookeeper/pull/451 bad news is that the review process over there is going just as well as this one. Alternatively, if you want to take an extreme approach, here's a fork of Kafka backed by ETCD instead of Zookeeper: https://github.com/banzaicloud/apache-kafka-on-k8s

regispl:
@aermakov-zalando I think this PR would benefit few other people running Zookeeper for Kafka, so they can workaround the issue with IP caching mentioned by @imduffy15 in the description (not sure if all the people are aware of the problem and possible workaround though - some of them are for sure 😉 ). 

Is there any guide on how to "release" my "local" Taupage, so it's available in live environment? I can see the "Build a new AMI" section in README, but how do I make it available in live environment? Or is this command publishing it as well?

Anyway, I don't have time to do it right now, but if there's a light at the end of the tunnel, that this PR could solve our problem, I think it would be easier for me to get that prioritised sooner than later.

@imduffy15 Thanks for update! 😃

aermakov-zalando:
I'm not sure it'll easy enough for you to build a new Taupage image, but we can setup a job to build an image from a PR. Please ping us when you have time, then.

PR #480, opened 2018-01-08
update docker

drummerwolli:
this PR mostly serves as a conversation starter, since i was not able to build taupage by myself currently and therefore cannot check this really. please check the feasibility of updating docker. the version currently used in taupage is almost a year old.

### DIFF
--- a/build/setup.d/20-install-packages.sh
+++ b/build/setup.d/20-install-packages.sh
@@ -6,7 +6,7 @@
 pkgs="
 auditd
 build-essential
-docker-engine=1.12.6-0~ubuntu-trusty
+docker-ce=17.12.0~ce-0~ubuntu-trusty
 gcc
 htop
 iproute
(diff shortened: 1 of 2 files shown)


mikkeloscar:
@drummerwolli Thanks for opening the PR and starting the conversation!

Will updating docker provide new features you need or fix bugs present in the current version?

I ask because we have bad experience with many docker versions especially in our Kubernetes infrastructure (where docker is exercised more than in taupage). So unless there are problems with the current version we don't see the urgency to update what is, from our PoV, a stable docker version.

drummerwolli:
Hi, Mikkel, this would be clearly to introduce new features. specifically this: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact which was introduced with https://github.com/moby/moby/pull/31352 in V17.05-ce (https://docs.docker.com/release-notes/docker-ce/#17050-ce-2017-05-04, first paragraph, second bullet point)

i see that some testing is definately needed, but wouldn't it be even for the Kubernetes Infrastructure also be beneficial, if not at some point crucial to update docker as well?

hjacobs:
@drummerwolli the Kubernetes infrastructure uses CoreOS ContainerLinux as the base AMI. I'm closing this PR for now as we will not invest time to test and roll out a new Docker version.

PR #481, opened 2018-01-11
Configuration parameter to disable taking process snapshots on Nodejs

vibhory2j:
A memory leak issue was reported in NodeJs application agent for which appdynamics has suggested to reduce process snapshots to 0 explicitly in agent configuration. This pull request is raised to add a config parameter (NODEJS_SNAPSHOT=$config_nodejs_disable_snapshots) in senza yaml. 

Tested it locally in a stubbed environment and the code snippet updates the config.json properly.

Can we release the AMI to only monitoring-appdynamics account before releasing it to the effected team(s).


aermakov-zalando:
I'd appreciate if you refactored the script to use something like `jq` instead of modifying JSON with `sed`, but this looks good.

vibhory2j:
👍

PR #484, opened 2018-01-26
Fix nvidia checksum validation

aermakov-zalando:
The checksum validation code used `-ne` instead of `!=` and because the hashes couldn't be parsed as numbers it always worked no matter what. Fix it so it actually validates something.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #485, opened 2018-01-26
initramfs: add nvme support for booting on c5/m5 instances

### LINKED ISSUE #482: m5 and c5 instance support
They are not only slightly cheaper than previous generation but also showing better cpu performance.
I think it should be enough just to change base ubuntu ami to get it working (the current bas ami is rather old)

aermakov-zalando:
m5/c5 instances weren't supported because the rather old initramfs configuration we have didn't have the `nvme` module and m5/c5 use NVMe for the root volumes as well. Fix #482.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #486, opened 2018-01-31
Use new channel-based release process

aermakov-zalando:
New AMIs are now built as `TaupageBuild-<NUMBER>`. The sharing script now accepts a `channel` argument, and creates a copy of the AMI with the new name (`Taupage<CHANNEL>-AMI-<PUBLISH_DATE>`). Original build date is stored in a tag for use in reporting and compliance checks.


aermakov-zalando:
👍

mikkeloscar:
:+1:

Issue #489, opened 2018-03-16
Propagate $ENABLE_INTROSPECTION environment variable passthough to the planb-tokeninfo fix


smirnov:
Enable passthrough of environment variable ENABLE_INTROSPECTION to the planb upstart job and startup script.

jbspeakr:
seems to work with TaupageStaging-AMI-20180323-150822. close?

jbspeakr:
wait. this looks weird: 
- in the [taupage setup](https://github.com/zalando-stups/taupage/blob/master/build/setup.d/55-setup-taupage-yaml.sh#L39) the variable ´tokeninfo_enable_introspection´ get's written into meta. - but in [tokeninfo init](https://github.com/zalando-stups/taupage/blob/master/runtime/etc/init/planb-tokeninfo.conf#L17) it expects `tokeninfo_introspection` to be present in meta. 

I would expect introspection not to work anymore after a reboot. also those inconsistencies make it really hard to figure out what to use. please make it consistent to [our current docs](https://pages.github.bus.zalan.do/foundation/docs/oauth2-guide/tokeninfo-introspection.html#validating-tokens-using-introspection-endpoint).

jbspeakr:
also it would  be nice to have a consistent naming schema for related config:
`local_planb_tokeninfo: true` together with `tokeninfo_introspection: true` is not nice.

I would prefer something like  `local_planb_tokeninfo: true`and `local_planb_tokeninfo_introspection: true`

smirnov:
The second instance of propagation (in the setup script, it's needed to propagate secret-vars.sh overlay) can be dropped altogether and just be kept in the upstart init script (then this setting will require to be set explicitly in senza.yaml).

PR #490, opened 2018-03-16
#489 allow enabling introspection for planb-tokeninfo service

smirnov:
Local planb-tokeninfo service takes into account ENABLE_INTROSPECTION environment variable. However currently there is no way for users to override it (taupage config variable doesn't get propagated to the startup script).

### DIFF
--- a/build/setup.d/55-setup-taupage-yaml.sh
+++ b/build/setup.d/55-setup-taupage-yaml.sh
@@ -35,3 +35,6 @@ fi
 if [ -n "$TOKENINFO_BUSINESS_PARTNERS" ]; then
     echo "tokeninfo_business_partners: $TOKENINFO_BUSINESS_PARTNERS" >> /meta/taupage.yaml
 fi
+if [ -n "$TOKENINFO_ENABLE_INTROSPECTION" ]; then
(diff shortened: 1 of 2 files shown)


aermakov-zalando:
Do you intend to have default values for this, or is this something that will only be configured per-stack?

smirnov:
This is something that should be possible for end-users to enable in the ENVIRONMENT section of the Taupage config, so that it propagates to tokeninfo on start.

aermakov-zalando:
The question was more about whether we should enable it by default or not.

smirnov:
Default should be false, with ability to override it to true.

aermakov-zalando:
LGTM then. Would you mind testing this with the Dev taupage channel once we merge this, with all 3 possible values (missing/false/true)?

szuecs:
:+1:

aermakov-zalando:
👍

smirnov:
:-1:

smirnov:
PR implementation will break existing apps, need to implement proper validation before propagation.

smirnov:
Tested by manually applying changes to running image. Seems to work just fine.

Will test dev image after merge.

smirnov:
:+1:

aermakov-zalando:
👍

PR #491, opened 2018-03-20
Fix share-ami.sh so we don't have to use empty CHANNEL

aermakov-zalando:
This doesn't work with our build pipeline. :(


aermakov-zalando:
👍

PR #493, opened 2018-04-09
Update ntp.conf to use AWS Time Sync Service

### LINKED ISSUE #475: Add support to Amazon Time Sync Service
Hi,

My team is experience some NTP drifts with Taupage. This is specially harmful since we are using Cassandra and the writes that happened in the past are ignored by design. (Long history short, some writes are being lost)

The current workaround is restarting the ntp daemons through tmux ssh. It works but it's a manual process and error-prone.

AWS has released *Amazon Time Sync Service*:

> a time synchronization service delivered over Network Time Protocol (NTP) which uses a fleet of redundant satellite-connected and atomic clocks in each region to deliver a highly accurate reference clock

https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/

It looks really promising and it would help us to reduce the ntp deltas to the minimun.

aermakov-zalando:
Use the default config with only the servers changed.
Fix #475.

### DIFF
--- a/runtime/etc/ntp.conf
+++ b/runtime/etc/ntp.conf
@@ -1,0 +1,55 @@
+# /etc/ntp.conf, configuration for ntpd; see ntp.conf(5) for help
+
+driftfile /var/lib/ntp/ntp.drift
+
+
+# Enable this if you want statistics to be logged.
+#statsdir /var/log/ntpstats/
+
+statistics loopstats peerstats clockstats
+filegen loopstats file loopstats type day enable
+filegen peerstats file peerstats type day enable
+filegen clockstats file clockstats type day enable
+
+# Specify one or more NTP servers.
+
+# Use servers from the NTP Pool Project. Approved by Ubuntu Technical Board
+# on 2011-02-08 (LP: #104525). See http://www.pool.ntp.org/join.html for
+# more information.
+server 0.ubuntu.pool.ntp.org
+server 1.ubuntu.pool.ntp.org
+server 2.ubuntu.pool.ntp.org
+server 3.ubuntu.pool.ntp.org
+
+# Use Ubuntu's ntp server as a fallback.
+server ntp.ubuntu.com
+
+# Access control configuration; see /usr/share/doc/ntp-doc/html/accopt.html for
+# details.  The web page <http://support.ntp.org/bin/view/Support/AccessRestrictions>
+# might also be helpful.
+#
+# Note that "restrict" applies to both servers and clients, so a configuration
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

szuecs:
:+1:

runtime/etc/ntp.conf
```diff
@@ -0,0 +1,46 @@
...
+#statsdir /var/log/ntpstats/
+
+statistics loopstats peerstats clockstats
+filegen loopstats file loopstats type day enable
+filegen peerstats file peerstats type day enable
+filegen clockstats file clockstats type day enable
+
+# Specify one or more NTP servers.
+# Use AWS Time Sync Service
+server 169.254.169.123
```
Cesarla: @aermakov-zalando Despite using AWS Time Sync Service, we still observed drifting between instances freshly created, taking a long time to heal by themselves.

A freshly deploy instance:
```
remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
*******                      3 u   11   64   37    0.095  206.255 104.590
```

Same instance after 30 minutes:
```
remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
*******                      3 u   11   64  377    0.122   32.242  42.379
```

I have checked AWS documentation and suggests to use `prefer iburst` in order to force a fast synchronisation, why this option is not enabled?:
https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/

aermakov-zalando: Thanks for the suggestion, we'll enable `iburst` to ensure that the time syncs up more quickly after startup.

However, I would recommend against relying on the system clock being perfectly synchronised. AWS doesn't provide an SLA for the Time Sync Service or even any guarantees about the clock precision, you could run into issues with ntpd (which nobody actually monitors), and your application might still generate slightly incorrect timestamps due to GC pauses, network failures or other instance-related issues.

aermakov-zalando: PR: https://github.com/zalando-stups/taupage/pull/498

Cesarla: @aermakov-zalando Kudos for the quick action!!

Unfortunately, we have to relay in the wall clock since my team is heavily invested in Cassandra and e.g: [Why Cassandra cluster need synchronized clocks between nodes?](https://stackoverflow.com/questions/34898693/why-cassandra-cluster-need-synchronized-clocks-between-nodes)

The time drifts causes to lose some writes and eventually has customer impact.

PR #495, opened 2018-05-03
Moved common logging attributes to server_attributes

beverage:
Per-log attributes are counted as log volume per log line.

### DIFF
--- a/runtime/opt/taupage/init.d/01-register-scalyr-agent.sh
+++ b/runtime/opt/taupage/init.d/01-register-scalyr-agent.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 # read /meta/taupage.yaml
 # get scalyr Key and register the agent
-
 #read taupage.yaml file
 eval $(/opt/taupage/bin/parse-yaml.py /meta/taupage.yaml "config")
-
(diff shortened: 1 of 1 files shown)


femueller:
:+1:

aermakov-zalando:
Can you add AWS account ID and region if you're modifying this anyway? They should be available from the instance identity document at http://169.254.169.254/latest/dynamic/instance-identity/document.

aermakov-zalando:
And please consider making the script readable with the help of `jq` or `python`. :)

aermakov-zalando:
What happens if an instance is terminated and then another one is started with a different stack but the same IP address as the previous one? Will the attributes be overwritten completely or will I have two different servers in the list?

beverage:
You will have the same serverHost as taupage sets it to the application id.  You will get a second stack name in the list.  ServerHost is something that should probably never be queried on for this reason - I am not sure why, long ago, it was made redundant with application_id.

femueller:
👍

femueller:
:+1:

beverage:
:+1:

aermakov-zalando:
👍

PR #496, opened 2018-05-04
syslog: optionally add AWS metadata to the logs

aermakov-zalando:
If `rsyslog_aws_metadata` is set in Taupage config, add AWS account ID and region to the log lines. This is useful for Scalyr S3 archival, because it only exports the raw log lines which means it's impossible to determine which account the message came from.


szuecs:
:+1:

szuecs:
:+1:

aermakov-zalando:
👍

PR #497, opened 2018-05-08
#212 prohibit latest and SNAPSHOT Docker image tags

### LINKED ISSUE #212: Prohibit using mutable *-SNAPSHOT Docker images
I don't think we should remove the SNAPSHOT feature from Pier One (https://github.com/zalando-stups/pierone/issues/74), rather, reject them with Taupage (abort startup) then we still keep the sharing between team members.

hjacobs:
Disallow mutable Docker image tags such as `latest` or `*-SNAPSHOT`.

Fixes #212.

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -356,12 +356,47 @@ def wait_for_health_check(config: dict):
     sys.exit(2)
 
 
+def is_valid_source(source):
+    '''
+    >>> is_valid_source('')
+    False
+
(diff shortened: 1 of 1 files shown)


aermakov-zalando:
👍

mikkeloscar:
:+1:

aermakov-zalando:
Do we want to ban other registries as well, by the way? Or do we want to still rely on fullstop for that?

hjacobs:
@aermakov-zalando yeah, I guess we could whitelist registries, but that needs some configuration option. I went with this simple first step.

PR #498, opened 2018-05-14
ntpd: enable iburst

aermakov-zalando:
This leads to time being synced very quickly after boot instead of only after several minutes.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #499, opened 2018-05-15
Don't send ntp traffic via NAT-GW

CyberDem0n:
Commit https://github.com/zalando-stups/taupage/commit/36907ee43133cc49d068efbad19737ab58a95077 introduced `AWS Time Sync Service` running on 169.254.169.123, which is ip of a hypervisor.
It broke DMZ deployments without public ip.


szuecs:
:+1:

CyberDem0n:
:+1:

PR #501, opened 2018-06-06
Enforce Logging to eu.scalyr.com

### LINKED ISSUE #500: Enforce eu region for Scalyr
We need to implement logging by default only to eu.scalyr.com.

femueller:
This fixes #500.


szuecs:
:+1:

PR #502, opened 2018-07-31
Implement trusted image validation

aermakov-zalando:
Add a simple check if the image is trusted in the Docker registry before running it. This is done by parsing the image tag and checking for the `X-Trusted: true` header in the image manifest response. Note that this only applies to the "main" container that's started by Taupage, enforcing it for all pulls would break a lot of things like CI containers.
An additional benefit is that this change would also make it impossible to run images from registries other than Pier One since they don't send the header in the first place. This is already a compliance requirement, but there's a bit more enforcement now.

### DIFF
--- a/runtime/opt/taupage/runtime/Docker.py
+++ b/runtime/opt/taupage/runtime/Docker.py
@@ -303,32 +303,12 @@ def get_other_options(config: dict):
         yield '--shm-size={}'.format(config.get('shm_size'))
 
 
-def registry_requires_auth(registry: str):
-    return registry == 'pierone.stups.zalan.do'
-
-
 def registry_login(config: dict, registry: str):
-    if registry_requires_auth(registry):
-        pierone_url = 'https://{}'.format(registry)
-        pierone.api.docker_login_with_iid(pierone_url)
-
-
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/runtime/Docker.py
```diff
@@ -356,46 +376,75 @@ def wait_for_health_check(config: dict):
...
+    ...
+    ValueError: Image tag not specified
 
+    >>> parse_image_tag('registry.example.org/foo/nginx:')
+    Traceback (most recent call last):
+    ...
+    ValueError: Image tag not specified
 
-def main(args):
+    >>> parse_image_tag('registry.example.org/foo/nginx:latest')
```
linki: Parsing a string into an image dict seems to also exist in the PierOne API client code: https://github.com/zalando-stups/pierone-cli/blob/master/pierone/api.py#L27

linki:
Looks good but there's some duplicated code that seems already present in PierOne's api client, e.g. parsing an image from a string and getting the value of the trusted attribute: https://github.com/zalando-stups/pierone-cli/blob/master/pierone/api.py

mikkeloscar:
:+1:

szuecs:
:+1:

PR #508, opened 2018-08-27
Increase number of attempts to find EBS volume

CyberDem0n:
If we are running in ASG and instance got unhealthy, ASG will first remove it from the group, spawn a replacement. The old instance will be terminated in only in about 10 minutes.
As a result, the new instance is failing to attach EBS volumes because they are still attached to the old instance.
In order to fix this problem, I increase the number of attempts to find the volumes to 10 and increase the timeout between attempts to 2 minutes.

Hope it should be enough to cover 99% of failure cases.


Jan-M:
:+1:

Jan-M:
Sounds good, and probably also benefits others with EBS volumes dangling around.

avaczi:
:+1:

PR #510, opened 2018-09-24
Include Fluentd support

femueller:
Signed-off-by: Felix Mueller felix.mueller@zalando.de

This PR will add support for Fluentd as an alternative distributed logging agent.

Backward compatible, the Scalyr configuration up until now is still recognized and Scalyr agent will still be used, Fluentd will not be run in that case.

With Fluentd the following destinations for logs are available as of now: scalyr, s3, rsyslog


femueller:
Signed-off-by: Felix Mueller felix.mueller@zalando.de

PR #511, opened 2018-09-24
Add optional Fluentd support

femueller:
Signed-off-by: Felix Mueller <felix.mueller@zalando.de>

This PR will add support for Fluentd as an alternative distributed logging agent.

Backward compatible, the Scalyr configuration up until now is still recognized and Scalyr agent will still be used, Fluentd will not be run in that case.

With Fluentd the following destinations for logs are available as of now: scalyr, s3, rsyslog


build/setup.d/21-install-td-agent-packages.sh
```diff
@@ -0,0 +1,14 @@
+# Ensure that /etc/init.d/td-agent is removed
+rm /etc/init.d/td-agent
+
+# Ensure that directories are writable
+chmod -R 1777 /tmp/
```
aermakov-zalando: Why is this needed?

build/setup.d/21-install-td-agent-packages.sh
```diff
@@ -0,0 +1,14 @@
+# Ensure that /etc/init.d/td-agent is removed
```
aermakov-zalando: Please add at least a shebang. Ideally the script should also fail if something inside fails (even though taupage-init currently does nothing) so we can improve it later.

runtime/etc/init/td-agent.conf
```diff
@@ -0,0 +1,11 @@
...
+
+stop on shutdown
+
+respawn
+
+pre-start script
+    mkdir -p -m0755 /var/run/td-agent
+end script
+
+exec /opt/td-agent/embedded/bin/ruby -Eascii-8bit:ascii-8bit /usr/sbin/td-agent --log /var/log/td-agent/td-agent.log --daemon /var/run/td-agent/td-agent.pid --no-supervisor
```
aermakov-zalando: Who creates `/var/log/td-agent`?

runtime/opt/taupage/init.d/02-prepare-td-agent.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Ensure that directories are writable
+chmod -R 1777 /tmp/
```
aermakov-zalando: Again, why?

runtime/opt/taupage/init.d/02-register-td-agent.py
```diff
@@ -0,0 +1,142 @@
...
+        # If scalyr_api_key starts with "aws:kms:" then decrypt key
+        match_kms_key = re.search('aws:kms:', scalyr_api_key, re.IGNORECASE)
+        if match_kms_key:
+            scalyr_api_key = re.sub(r'aws:kms:', '', scalyr_api_key)
+            try:
+                scalyr_api_key = subprocess.check_output(['python3',
+                                                          '/opt/taupage/bin/decrypt-kms.py',
+                                                          scalyr_api_key]).decode('UTF-8').strip()
+            except Exception:
+                logger.error('Failed to run /opt/taupage/bin/decrypt-kms.py')
```
aermakov-zalando: Why not just crash here? What's the point of proceeding?

build/setup.d/20-install-packages.sh
```diff
@@ -33,7 +33,9 @@ rkhunter
 rsyslog-gnutls
 ruby
 scalyr-agent-2
+supervisor
```
aermakov-zalando: Why? It doesn't seem to be used.

runtime/opt/taupage/init.d/02-prepare-td-agent.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
gargravarr: #!/usr/bin/env bash

runtime/opt/taupage/init.d/02-register-td-agent.py
vwiessner: There could be addititional destinations whose configuration depends on this script finishing, which should still work if the Scalyr key decryption fails.

runtime/opt/taupage/init.d/02-prepare-td-agent.sh
femueller: Why? We use `#!/bin/bash` for most scripts.

build/setup.d/21-install-td-agent-packages.sh
femueller: I can't remember the exact reason. I will revert this change and see if we still require it.

runtime/opt/taupage/init.d/02-prepare-td-agent.sh
femueller: Same as above https://github.com/zalando-stups/taupage/pull/511#discussion_r220101407.

build/setup.d/21-install-td-agent-packages.sh
femueller: Will do.

build/setup.d/21-install-td-agent-packages.sh
femueller: This doesn't appear to be necessary.

runtime/opt/taupage/init.d/02-prepare-td-agent.sh
femueller: This doesn't appear to be necessary anymore.

runtime/opt/taupage/init.d/02-register-td-agent.py
aermakov-zalando: But then the config will be invalid since you don't quote the api key in the template. If you want to rely on this, please test the behavior.

build/setup.d/21-install-td-agent-packages.sh
aermakov-zalando: The script still doesn't fail.

femueller:
@vwiessner and me have submitted the requested changes. Can you please check again?

aermakov-zalando:
👍

vwiessner:
:+1:

aermakov-zalando:
👎

aermakov-zalando:
Oops. I've just noticed that you fixed Travis in the PR as well. Do you mind removing the fix from here since it's fixed in `master`?

vwiessner:
:+1:

aermakov-zalando:
👍

PR #518, opened 2018-10-29
Adding dummy Scalyr Key for test

femueller:
In order to properly test https://github.com/zalando-stups/taupage/blob/master/tests/spec/localhost/scalyr-agent_spec.rb a (dummy) key actually needs to be configured.

Signed-off-by: Felix Mueller <felix.mueller@zalando.de>

### DIFF
--- a/test-userdata.yaml
+++ b/test-userdata.yaml
@@ -83,4 +83,4 @@ sysfs:
   /sys/kernel/mm/transparent_hugepage/enabled: never
 
 # Dummy Scalyr key
-scalyr_account_key: 123
+scalyr_account_key: foo1234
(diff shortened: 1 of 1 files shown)


test-userdata.yaml
```diff
@@ -81,3 +81,6 @@ sysctl:
 # Edit sysfs paths
 sysfs:
   /sys/kernel/mm/transparent_hugepage/enabled: never
+
+# Dummy Scalyr key
+scalyr_account_key: 123
```
mikkeloscar: Should the key be an int? It seems wrong. I would assume it should be a string?

femueller: Updated it to a string. But it shouldn't really have an effect in any way on the test.
The test itself just checked if scalyr_key is configured and checks if the config is properly set.

mikkeloscar:
:+1:

femueller:
👍

PR #522, opened 2018-10-30
Handled ENV variable which is non-string.

arjunrn:
Added test also.

Signed-off-by: Arjun Naik <arjun.naik@zalando.de>


aermakov-zalando:
👍

PR #526, opened 2018-11-28
Use trusted test images

aermakov-zalando:
Use a trusted Docker image for the test instances so they can actually start.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #527, opened 2018-11-30
Handle environment config which is none

arjunrn:
If the `environment` config is set to `null` then the check for secret environment variables fails. This PR fixes that edge case.

Signed-off-by: Arjun Naik <arjun.rn@gmail.com>


lotharschulz:
👍

Issue #529, opened 2018-12-18
Fluentd agent uses port 8888


arjunrn:
The `fluentd` agent listens on port _8888_ which collides with any docker container also using that port. The port should be disabled or changed to listen on another port.

PR #531, opened 2018-12-18
added command to switch off td-agent by default

vwiessner:
fix for issue #529

Signed-off-by: Veit Wiessner <veit.wiessner@zalando.de>

### DIFF
--- a/runtime/opt/taupage/init.d/02-prepare-td-agent.sh
+++ b/runtime/opt/taupage/init.d/02-prepare-td-agent.sh
@@ -4,3 +4,5 @@
 mkdir -p -m0755 /var/run/td-agent
 
 service td-agent start
+sleep 5
+service td-agent stop


arjunrn:
Why start and stop the agent? Why not just don't start the agent.

aermakov-zalando:
I agree with @arjunrn, seems pretty stupid to start the agent and immediately stop it back. I'm also not sure if this will work correctly after a reboot, because IIRC init scripts are only ran once.

vwiessner:
Good thing github mails the first version of every comment, not the edited one, I always like to know how people really feel.

I'll change it.

aermakov-zalando:
@vwiessner sorry, I didn't mean to offend!
This still won't work correctly, unless I'm misunderstanding something. `taupage-init` only runs once for every node, so either the agent would restart after a reboot when it shouldn't, or it wouldn't start up when it should.

vwiessner:
I'll tested it with fluentd enabled and disabled with a taupage image build from this branch
It's not running when disabled, even after reboot (sudo reboot)
It's runnig when enabled, even after reboot (sudo reboot)

arjunrn:
:+1:

vwiessner:
:+1:

PR #534, opened 2019-01-08
Fluentd extend s3 options

vwiessner:
made some more options for s3 plugin configurable from taupage.yaml


runtime/opt/taupage/init.d/02-register-td-agent.py
```diff
@@ -97,6 +99,8 @@ def update_configuration_from_template():
     # Get Scalyr key only if configured
     if fluentd_destinations.get('scalyr') or fluentd_destinations.get('scalyr_s3'):
         scalyr_api_key = get_scalyr_api_key()
+    else:
+        scalyr_api_key = False
```
aermakov-zalando: I'd set it to `None` to avoid confusion.

PR #536, opened 2019-01-16
Reduce rsyslogd output

femueller:
This Pull-Request by default reduces redundant log output by default which is logged by application logging libraries anyways such as the date.
It also removes the Docker container ID and instance hostname and moves both over to the Scalyr server fields.


aermakov-zalando:
👍

femueller:
👍

PR #537, opened 2019-01-16
added option to configure scalyr agent sampling rate

vwiessner:
Signed-off-by: Veit Wiessner <veit.wiessner@zalando.de>


vwiessner:
Changes to standard behavior:
* Standard Scalyr setup: none
* Fluentd: 
  * Scalyr and rsyslog: none
  * S3:
    * acl: default changed from ```private``` to ```bucket-owner-full-control```  (can be overridden with config)
    * timekey (length of time after which buffer is flushed): changed from 1m to 5m (can be overridden with config)
    * S3 target path changed to ```/taupage/<application>/<version>/<path>/<year>/<month>/<day>/<hour>/<hostname>/<timeslice>_<index>.gz```

vwiessner:
Fair point. However, I did not create this script, I merely work with whats already there. I agree that it should be refactored, however, right now I don't have time to do it but this change needs to be made.

Quick question: Will you block this PR if I don't change it?

vwiessner:
:+1:

christianberg:
👍

christianberg:
This setup has been tested by Keystone in their Staging environment and should get a staging release, so they can move on to test it in live.

I would like to look into two refactorings before we make further changes to this code:
- Replace the usage of `sed` in the bash scripts with `jq`, or move to a Python script.
- Reduce the amount of duplication we currently have in the fluentd configuration.

PR #540, opened 2019-02-08
textfile collector configuration

vibhory2j:
## FluentD Monitoring
We would like to monitor the _fluentd_ agent in conjunction with [ZMON](http://zmon.io) via the prometheus node exporter. This pull request provides configuration for textfile collector in node exporter and fluentd role for the machine. 

Needs to be tested before merging.


vwiessner:
duplicate of #541

PR #541, opened 2019-02-12
Fluentd prometheus

vwiessner:
Adds fluentd metrics to node_exporter output


runtime/opt/taupage/init.d/02-register-td-agent.py
```diff
@@ -152,6 +152,21 @@ def update_configuration_from_template():
     if not logging_config:
         logger.info('Found no logging section in senza.yaml')
         raise SystemExit()
+    else:
+        try:
```
arjunrn: using a file context manager is a better pattern

vwiessner: good point, changed

aermakov-zalando:
LGTM. Does it degrade gracefully if the log doesn't contain valid utf-8?

vwiessner:
@aermakov-zalando can you go into detail what you mean? Fluentd doesn't crash, but the non UTF8 will show up garbled in Scalyr or on S3.

aermakov-zalando:
That's fine, I just wanted to avoid a situation where Python would throw an encoding exception and not upload anything at all.

vwiessner:
:+1:

christianberg:
👍

PR #543, opened 2019-03-12
Update the Taupage base image to Ubuntu 16.04 LTS

arjunrn:
Upgrade the base AMI to Ubuntu 16.04

Changelog:

* Updated the `cloud-init` patch to work with the library based in Ubuntu 16.04
* Install `upstart` and purge `systemd`
* Updated OpenJDK on the base AMI to Java 8
* Removed `serverspec` tests which assume that Ubuntu 16.04 has systemd.

Build changes:
* Added tags so that build and test machines/volumes can be bulk deleted. If there are test machines which are left running this causes new tests to fail. The only way to run them reliability is to ensure that only set of tests is running at a time.
* Rebuilt the docker images used for smoke tests so that they rely on the correct base image libraries.


vwiessner:
:+1:

femueller:
👍

build/setup.d/10-upgrade-base.sh
```diff
@@ -5,6 +5,7 @@ set -x
 echo "Updating system..."
 
 # sudo rm -rf /var/lib/apt/lists/*
+chmod 777 /tmp
```
aermakov-zalando: Why?

aermakov-zalando:
LGTM, other than the permission change on `/tmp`. Is it really needed?

arjunrn:
> LGTM, other than the permission change on `/tmp`. Is it really needed?

For some reason after the reboot after installing `upstart` the permissions on the `/tmp` directory are:

```
drwxr-xr-x   3 root root  4096 Mar 25 14:27 tmp/
```
This is the [bug](https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1577926)

aermakov-zalando:
I see. Could you set it to `1777` then?

aermakov-zalando:
👍

arjunrn:
:+1:

PR #545, opened 2019-03-18
Upgrade Nvidia driver which is compatible with latest kernel release

arjunrn:
The current master cannot be built because of changes in the kernel of xenial. This driver version is the next version above and should be a significant change from the previous version.


arjunrn:
@vwiessner Here's the PR

vwiessner:
:+1:

aermakov-zalando:
👍

PR #547, opened 2019-04-01
Reverting the logging to S3 changes

arjunrn:
Temporarily revert the S3 changes so that the PR #543 can be rolled out.

### DIFF
--- a/build/setup.d/01-add-node_exporter-text_collector-dir.sh
+++ b/build/setup.d/01-add-node_exporter-text_collector-dir.sh
@@ -2,5 +2,4 @@
 
 mkdir -p -m0755 /var/local/textfile_collector
 
-echo "fluentd_process_running{tag=\"td-agent\",hostname=\"$(hostname)\"} 0.0" > /var/local/textfile_collector/fluentd_metrics.prom
-echo "fluentd_default_s3_logging{tag=\"td-agent\",hostname=\"$(hostname)\"} 0.0" > /var/local/textfile_collector/fluentd_default_s3.prom
+echo "fluentd{tag="fluentd_enabled"} 0.0" > /var/local/textfile_collector/fluentd_metrics.prom
--- a/build/setup.d/01-add-node_exporter-text_collector-dir.sh
+++ b/build/setup.d/01-add-node_exporter-text_collector-dir.sh
@@ -2,4 +2,4 @@
 
 mkdir -p -m0755 /var/local/textfile_collector
 
-echo "fluentd{tag="fluentd_enabled"} 0.0" > /var/local/textfile_collector/fluentd_metrics.prom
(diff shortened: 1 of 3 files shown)


build/setup.d/01-add-node_exporter-text_collector-dir.sh
```diff
@@ -2,5 +2,4 @@
 
 mkdir -p -m0755 /var/local/textfile_collector
 
-echo "fluentd_process_running{tag=\"td-agent\",hostname=\"$(hostname)\"} 0.0" > /var/local/textfile_collector/fluentd_metrics.prom
-echo "fluentd_default_s3_logging{tag=\"td-agent\",hostname=\"$(hostname)\"} 0.0" > /var/local/textfile_collector/fluentd_default_s3.prom
+echo "fluentd{tag="fluentd_enabled"} 0.0" > /var/local/textfile_collector/fluentd_metrics.prom
```
szuecs: please escape correctly

runtime/opt/taupage/init.d/02-register-td-agent.py
```diff
@@ -47,22 +47,18 @@ def get_scalyr_api_key():
                 raise SystemExit()
         if scalyr_api_key == "Invalid KMS key.":
             logger.error('Failed to decrypt KMS Key')
-            raise SystemExit(1)
+            raise SystemExit()
```
szuecs: why you change that?
 != 0 is an error and default might be return 0..

runtime/opt/taupage/init.d/02-register-td-agent.py
```diff
@@ -47,22 +47,18 @@ def get_scalyr_api_key():
...
         return scalyr_api_key
 
 
 def update_configuration_from_template():
     ''' Update Jinja Template to create configuration file for Scalyr '''
     fluentd_destinations = dict(scalyr=False, s3=False, rsyslog=False, scalyr_s3=False)
     config = get_config()
-    logging_config = config.get('logging', {})
+    logging_config = config.get('logging')
```
szuecs: This will crash in line 75 if unset

aermakov-zalando:
I agree with @szuecs's comments; it's unfortunate that we've missed this in the first place. We could keep these fixed, or do you want to keep it as a pure revert PR?

christianberg:
Thanks for the review. Let's keep the fixes @szuecs mentioned in. Would be silly to risk an NPE just to cleanly revert the PR.

arjunrn:
@aermakov-zalando @christianberg @szuecs I've added those changes back. Can you have a look again?

aermakov-zalando:
👍

arjunrn:
:+1:

PR #548, opened 2019-04-09
Mask JWT tokens for logs send by Scalyr Agent

christianberg:
Any strings matching the format of a [JWT](https://tools.ietf.org/html/rfc7519) are replaced with `+++JWT_TOKEN_REDACTED+++` before they are shipped Scalyr.


vwiessner:
:+1:

PR #549, opened 2019-04-10
Logging s3 by default

vwiessner:
Log to S3 by default
set up a cronjob to test s3 iam permissions if s3 was chosen as log destination


christianberg:
👍

vwiessner:
:+1:

PR #550, opened 2019-04-11
Install the GPU drivers from package manager

arjunrn:
Install the nvidia drivers from the Ubuntu package repositories to ensure compatibility.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -60,6 +60,75 @@ fi
 
 AMI_ID=$(aws ec2 describe-images --region $region --filters "Name=name,Values=TaupageBuild-$TAUPAGE_VERSION" --query 'Images[*].{ID:ImageId}' --output  text)
 
+
+echo "Starting gpu server..."
+result=$(aws ec2 run-instances \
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -60,6 +60,78 @@ fi
...
+    if [ $alive -eq 0 ]; then
+        break
+    fi
+
+    sleep 2
+done
+
+terminationcounter=300
+
+while [ true ]; do
```
szuecs: why not a for loop here?
`for i in {1..300}; do ...; done`

arjunrn: Because I just copied a similar test and made some small changes.

arjunrn: Also with a fixed number of attempts there will have to be error handling and in bash. This way the pipeline just times out. Also it's expected that this step won't fail.

szuecs: true, but with a for loop you do not need to use the bash math expression and reduce the lines of code.

aermakov-zalando:
👍

arjunrn:
:+1:

PR #553, opened 2019-04-25
Set THP to madvise in grub

arjunrn:
Set the THP to madvise to prevent overconsumption of memory in jobs with many forked processes.


njuettner:
👍

arjunrn:
:-1:

PR #555, opened 2019-05-21
Specifying role in cfn-signal command

pc-alves:
To prevent situations where the `cfn-signal` command fails on init when the call is done across stacks, the role is specified when executing `cfn-signal`.
The role that is used with the command needs to have the following policy attached:
```
{
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Action": [
                    "cloudformation:SignalResource",
                    "cloudformation:DescribeStackResource"
                ],
                "Resource": "*"
            }
        ]
    }
```

### DIFF
--- a/runtime/opt/taupage/init.sh
+++ b/runtime/opt/taupage/init.sh
@@ -80,7 +80,7 @@ else
     echo "INFO: Notifying CloudFormation (region: $EC2_REGION, stack: $config_notify_cfn_stack, resource: $config_notify_cfn_resource, status: $result, IAM Role: $IAM_ROLE)..."
 
     ## First tries signal command with the role specified, and if that fails executes the signal command without the role
-    cfn-signal -e $result --role $IAM_ROLE --stack $config_notify_cfn_stack --resource $config_notify_cfn_resource --region $EC2_REGION || cfn-signal -e $result --stack $config_notify_cfn_stack --resource $config_notify_cfn_resource --region $EC2_REGION
+    cfn-signal -e "$result" --role "$IAM_ROLE" --stack "$config_notify_cfn_stack" --resource "$config_notify_cfn_resource" --region "$EC2_REGION" || cfn-signal -e "$result" --stack "$config_notify_cfn_stack" --resource "$config_notify_cfn_resource" --region "$EC2_REGION"
 fi
 
 END_TIME=$(date +"%s")
--- a/runtime/opt/taupage/init.sh
+++ b/runtime/opt/taupage/init.sh
@@ -75,9 +75,12 @@ else
     EC2_AVAIL_ZONE=$( curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone )
     EC2_REGION="`echo \"$EC2_AVAIL_ZONE\" | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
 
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/init.sh
```diff
@@ -75,9 +75,12 @@ else
     EC2_AVAIL_ZONE=$( curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone )
     EC2_REGION="`echo \"$EC2_AVAIL_ZONE\" | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
 
-    echo "INFO: Notifying CloudFormation (region: $EC2_REGION, stack: $config_notify_cfn_stack, resource: $config_notify_cfn_resource, status: $result)..."
+    IAM_ROLE=$( curl -s http://169.254.169.254/latest/meta-data/iam/security-credentials )
```
aermakov-zalando: This needs `--fail` and exit status checks.

runtime/opt/taupage/init.sh
```diff
@@ -75,9 +75,12 @@ else
...
 
-    cfn-signal -e $result --stack $config_notify_cfn_stack --resource $config_notify_cfn_resource --region $EC2_REGION
+    echo "INFO: Notifying CloudFormation (region: $EC2_REGION, stack: $config_notify_cfn_stack, resource: $config_notify_cfn_resource, status: $result, IAM Role: $IAM_ROLE)..."
+
+    ## First tries signal command with the role specified, and if that fails executes the signal command without the role
+    cfn-signal -e $result --role $IAM_ROLE --stack $config_notify_cfn_stack --resource $config_notify_cfn_resource --region $EC2_REGION || cfn-signal -e $result --stack $config_notify_cfn_stack --resource $config_notify_cfn_resource --region $EC2_REGION
```
aermakov-zalando: Quote the variables please, especially the role name.

runtime/opt/taupage/init.sh
lmineiro: It will, below the cfn-signal will short circuit and fallback to the current mode

aermakov-zalando: Yeah, but if the metadata API is unreachable I'd prefer a clear error message instead of having to figure out why `cfn-signal` failed.

lmineiro: You mean, like we've been doing on line 75 ;-)

lmineiro: Apart from the calls to the metadata endpoint, do you suggest any other improvements to the script?

runtime/opt/taupage/init.sh
pc-alves: Done

runtime/opt/taupage/init.sh
aermakov-zalando: Right, the whole script is pretty bad. :(

aermakov-zalando: Please sign the commits so the DCO check would pass.

aermakov-zalando:
👍

aermakov-zalando:
Please test both scenarios (role + no role) with the `dev` AMI after it's merged.

lmineiro:
:+1:

PR #558, opened 2019-05-23
Fixing misleading character

pc-alves:
> If it looks like a space, but it breaks your script, it is not a space.

In line 83, when the second operand of the OR operation is executed, the script fails with the following error message:
`taupage-init: /opt/taupage/init.sh: 83: /opt/taupage/init.sh:  cfn-signal: not found`

After multiple experiments, and hints from hard earned experienced, we identified a character in the second command that, despite every editor representing it as a blank space, it really isn't.
It is `0xc2a0` which in UTF-8 means a `NO-BREAK SPACE`.

Ref: http://www.fileformat.info/info/unicode/char/00a0/index.htm


lmineiro:
Who never... ;-)

lmineiro:
👍

aermakov-zalando:
👍

PR #560, opened 2019-05-27
Rolling back extra cfn-signal command on init script

pc-alves:
Tests on the base image revealed that this extra command is not necessary.
Rolling it back.


aermakov-zalando:
👍

lmineiro:
👍

PR #564, opened 2019-08-29
increase max line length to 50k for scalyr-agent logging

pheanex:
Increase max line length to 50k for scalyr agent


runtime/opt/taupage/init.d/01-register-scalyr-agent.sh
```diff
@@ -201,9 +201,29 @@ then
...
 fi
 
+echo -n "adding max_line_size... ";
+sed -i '/api_key/a \  max_line_size: 49900,' $scalyr_config
+if [ $? -eq 0 ];
+then
+    echo "DONE"
+else
+    echo "ERROR"
+    exit
```
vwiessner: you should always exit with !=0 in case of error, so the automated checks during taupage build will notice if it failed

pheanex: Good idea!

pheanex:
:+1:

njuettner:
You might want to run **shellcheck**. It seems there a lot of suggestions. Maybe we should also add `set -euo pipefail` in the script.

njuettner:
👍

PR #573, opened 2019-11-21
Correctly handle volume mappings on NVME-based instances

aermakov-zalando:
Use the scripts from [this repo](https://github.com/oogali/ebs-automatic-nvme-mapping) to correctly handle both old and new generation instances.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #574, opened 2019-11-21
Just use udev to manage symlinks

aermakov-zalando:
Let `udev` manage symlinks so dynamic attach/detach works.


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #576, opened 2020-01-15
Updated docker version

arjunrn:
The `docker-engine` package does not seem to be available anymore. Also it's extremely outdated.


aermakov-zalando:
👍

arjunrn:
Closing because docker restored the old version and I would rather not change anything in this undead monstrosity.

PR #578, opened 2020-02-04
Add open policy agent

amitrangra:
Add open policy agent download and install

### DIFF
--- a/build/setup.d/100-install-open-policy-agent.sh
+++ b/build/setup.d/100-install-open-policy-agent.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
+set -euo pipefail
 
-curl -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
+curl --fail -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
 mv opa /opt/taupage/bin
 chmod +x /opt/taupage/bin/opa
--- a/build/setup.d/100-install-open-policy-agent.sh
+++ b/build/setup.d/100-install-open-policy-agent.sh
@@ -2,3 +2,4 @@
 
 curl -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
 mv opa /opt/taupage/bin
+chmod +x /opt/taupage/bin/opa
--- a/build/setup.d/100-install-open-policy-agent.sh
+++ b/build/setup.d/100-install-open-policy-agent.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+curl -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
+mv opa /opt/taupage/bin
--- a/build/setup.d/100-install-open-policy-agent.sh
+++ b/build/setup.d/100-install-open-policy-agent.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-curl --fail -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
+curl --fail -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64 && echo "5e704decb04e8ef2963d2df209d84d3c42a73fc31c2ed61fd18b14296277411d opa" | sha256sum --check --status
 mv opa /opt/taupage/bin
 chmod +x /opt/taupage/bin/opa
--- a/runtime/etc/init/opa.conf
+++ b/runtime/etc/init/opa.conf
@@ -1,0 +1,27 @@
+description "Open Policy Agent"
+
+stop on shutdown
+
+respawn
+
+script
+      eval $(/opt/taupage/bin/parse-yaml.py /meta/taupage.yaml "config")
+      export AVAIL_ZONE="$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)"
+      export REGION="`echo \"$EC2_AVAIL_ZONE\" | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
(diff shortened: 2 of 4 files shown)


build/setup.d/100-install-open-policy-agent.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+curl -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
```
aermakov-zalando: Please use `--fail` and the usual boilerplate (`set -euo pipefail`) on top, and add a signature check. You'll also need to `chmod +x` it.

arjunrn:
Also move the upstart service here and corresponding code over here. Doesn't make sense to have it in 2 different places.

build/setup.d/100-install-open-policy-agent.sh
amitrangra: added the options

amitrangra:
> Also move the upstart service here and corresponding code over here. Doesn't make sense to have it in 2 different places.

Moved the upstart files here...

runtime/etc/init/opa.conf
```diff
@@ -0,0 +1,27 @@
+description "Open Policy Agent"
+
+stop on shutdown
+
+respawn
+
+script
+      eval $(/opt/taupage/bin/parse-yaml.py /meta/taupage.yaml "config")
+      export AVAIL_ZONE="$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)"
```
aermakov-zalando: This should also `--fail`, we've had some issues with IMDS availability in the past.

runtime/etc/init/opa.conf
```diff
@@ -0,0 +1,27 @@
...
+      export REGION="`echo \"$EC2_AVAIL_ZONE\" | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
```
aermakov-zalando: Please use `$()` instead of backticks, it makes it a bit easier to read. In general, fixing [shellcheck](https://www.shellcheck.net)'s complaints is a good idea when writing shell scripts.

runtime/etc/init/opa.conf
amitrangra: done

runtime/etc/init/opa.conf
amitrangra: changed

runtime/etc/init/opa.conf
```diff
@@ -0,0 +1,27 @@
+description "Open Policy Agent"
+
+stop on shutdown
+
+respawn
+
+script
+      eval $(/opt/taupage/bin/parse-yaml.py /meta/taupage.yaml "config")
+      export AVAIL_ZONE="$(curl --fail -s http://169.254.169.254/latest/meta-data/placement/availability-zone)"
+      export REGION="$(echo \"$EC2_AVAIL_ZONE\" | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:')"
```
aermakov-zalando: `$AVAIL_ZONE`. You also don't need to escape the quote characters, this will make your YAML invalid.

amitrangra: actually we referred https://github.com/zalando-stups/taupage/blob/master/runtime/opt/taupage/init.sh#L76

aermakov-zalando: I see. That code was unfortunately not updated, but I don't think we should use legacy hard to read features (such as backticks) for new code.

aermakov-zalando:
👍

build/setup.d/100-install-open-policy-agent.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+curl --fail -L -o opa https://github.com/open-policy-agent/opa/releases/download/v0.16.2/opa_linux_amd64
+mv opa /opt/taupage/bin
```
arjunrn: Can you verify the checksum of the downloaded file before proceeding?

runtime/etc/opa/opa_conf.yaml
```diff
@@ -0,0 +1,20 @@
+services:
+  - name: "${OPA_SERVICE_NAME}"
+    url: "${OPA_SERVICE_URL}"
+    credentials:
+      s3_signing:
+        metadata_credentials:
+          iam_role: "app-${APPLICATION_ID}"
```
arjunrn: This role might not always exist, because sometimes users create their own roles in the senza manifest.

arjunrn: On second thoughts, the instance would use the credentials from the metadata endpoint and it would have an existing role associated with it. As long as that role has permissions to read from the s3 bucket this line can be dropped.

build/setup.d/100-install-open-policy-agent.sh
amitrangra: Can you please share an example

runtime/etc/opa/opa_conf.yaml
amitrangra: opa documentation specify to provide the iam_role and region: https://www.openpolicyagent.org/docs/latest/configuration/#aws-signature
`To use the EC2 metadata service, the IAM role to use and the AWS region for the resource must both be specified as iam_role and aws_region respectively.`

build/setup.d/100-install-open-policy-agent.sh
jrake-revelant: https://github.com/zalando-stups/taupage/blob/master/build/setup.d/19-install-ec2-instance-connect.sh

runtime/etc/opa/opa_conf.yaml
arjunrn: Well then my first point holds. This is not the role which will always be used/present.

amitrangra: Agree but then to use opa, the intended applications would have to have a role like this...

arjunrn:
:+1:

aermakov-zalando:
👍

PR #579, opened 2020-02-06
Run the setup.d scripts in separate shells

aermakov-zalando:
Turns out the build scripts all run in the same shell for whatever reason. Adding a `set -euo pipefail` in the OPA install script broke some other scripts. Let's try fixing this mess.


aermakov-zalando:
👍

arjunrn:
:+1:

PR #584, opened 2020-02-19
CIAM-389: Start opa with enabled authorization

nuwandiw:
Signed-off-by: Nuwandi <nuwandi.wickramasinghe@zalando.de>

### DIFF
--- a/runtime/etc/init/opa.conf
+++ b/runtime/etc/init/opa.conf
@@ -24,11 +24,6 @@ script
       else
             export OPA_IAM_ROLE="$config_opa_iam_role"
       fi
-      if [ -z "$config_kio_app_id" ]; then
-            export KIO_APP_ID="$config_application_id"
-      else
-            export KIO_APP_ID="$config_kio_app_id"
-      fi
       if [ -z "$config_opa_service_name" ]; then
             export OPA_SERVICE_NAME="zalando"
       else
--- a/runtime/etc/init/opa.conf
+++ b/runtime/etc/init/opa.conf
@@ -20,7 +20,7 @@ script
             export OPA_SERVICE_URL="$config_opa_service_url"
       fi
       if [ -z "$config_opa_iam_role" ]; then
-            export OPA_IAM_ROLE=""
+            export OPA_IAM_ROLE="app-$config_application_id"
       else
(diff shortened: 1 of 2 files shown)


runtime/etc/opa/opa_conf.yaml
```diff
@@ -4,13 +4,13 @@ services:
...
         metadata_credentials:
-          iam_role: "app-${APPLICATION_ID}"
+          iam_role: "${OPA_IAM_ROLE}"
           aws_region: "${REGION}"
 
 labels:
   cloud_provider: aws
   execution_environment: "${OPA_EXECUTION_ENV}"
-  kio_app_ids: "${APPLICATION_ID}"
+  kio_app_ids: "${KIO_APP_ID}"
```
aermakov-zalando: Why does it need to be different from `APPLICATION_ID`?

nuwandiw: Since there was a concern with the role not always following the pattern app-${APPLICATION_ID}, we thought of being flexible with kio_app_id too. Also in general case nobody has to configure this and it will default to ${APPLICATION_ID}

But if you think, there cannot be a case where an application is using an `APPLICATION_ID` different to their kio app id, I would like to revert this line.

aermakov-zalando: I don't see any valid use cases where an instance would have multiple application identities. Let's just use the common configuration item for this and only make it configurable if it's really necessary.

runtime/etc/init/opa.conf
```diff
@@ -19,10 +19,20 @@ script
       else
             export OPA_SERVICE_URL="$config_opa_service_url"
       fi
+      if [ -z "$config_opa_iam_role" ]; then
+            export OPA_IAM_ROLE="app-$config_application_id"
+      else
+            export OPA_IAM_ROLE="$config_opa_iam_role"
+      fi
+      if [ -z "$config_kio_app_id" ]; then
```
aermakov-zalando: This isn't needed anymore, right?

nuwandiw: yes. Removed

aermakov-zalando:
👍

arjunrn:
:+1:

PR #585, opened 2020-02-19
install-cloudwatch-logs-agent: update pip first

aermakov-zalando:
It looks like the ancient `pip` we have right now grabs broken versions of some of the dependencies, let's update it.


aermakov-zalando:
👍

jrake-revelant:
👍

PR #586, opened 2020-02-21
Updated docker version

arjunrn:
The repository from where docker is downloaded will be [shutdown soon](https://github.com/moby/moby/issues/40466). Better upgrade now than later.


aermakov-zalando:
Is the `nvidia-docker` version compatible, or do we need to pin it as well?

aermakov-zalando:
👍

arjunrn:
@aermakov-zalando Don't know about the `nvidia-docker` but we have an e2e test now and it will fail if it's not compatible.

arjunrn:
:+1:

PR #589, opened 2020-03-19
Allow IPv4 address substitutions in EBS volume names

### LINKED ISSUE #488: Add support for variable expansion from the user data
To the best of my knowledge, the user data is static and can be found under /etc/taupage.yaml

Some use cases could benefit from having some basic variable expansion, based on a limited, well-known set of variables.

For [PlanB Cassandra](https://github.com/zalando-stups/planb-cassandra) each node has its own unique User Data but the only difference is the EBS volume name, for example:
```
volumes:
  ebs: {/dev/xvdf: cassandra-172.31.128.40}
```

Some tools that perform EC2 orchestration would benefit if the User Data was just a template and each instance would expand the known variables. For the above example, this would be something like:
```
volumes:
  ebs: {/dev/xvdf: cassandra-${local-ipv4}}
```

If this sounds useful, I propose an initial implementation with the following variables, directly from the metadata service - http://169.254.169.254/latest/meta-data/:
* local-ipv4
* hostname
* public-ipv4
* public-hostname

This would go into the get_config() method in https://github.com/zalando-stups/taupage/blob/master/runtime/usr/local/lib/python3.4/dist-packages/taupage/__init__.py providing the expansion for all calls already depending on this method. Any other requirements, like having access to the raw file, could still read the file /etc/taupage.yaml.

a1exsh:
Fixes #488

Limit scope of substitutions to EBS volume names only.  This is not needed in
any other places of configuration (otherwise there would be open issues for
those as well).

Only allow `local-ipv4` and `public-ipv4` vars from instance metadata: for the
volume name substitutions these are the most useful ones.

### DIFF
--- a/runtime/opt/taupage/init.d/10-prepare-disks.py
+++ b/runtime/opt/taupage/init.d/10-prepare-disks.py
@@ -2,6 +2,7 @@
 
 import argparse
 import logging
+import string
 import sys
 import subprocess
 import os
@@ -28,6 +29,29 @@ def detect_region():
 def zone():
     """Helper to return the AZ for the current instance"""
     return boto.utils.get_instance_metadata()['placement']['availability-zone']
+
+
+def select_keys(d, keys):
+    """
+    >>> select_keys({'foo': 1, 'bar': 2, 'baz': 'def'}, ['foo', 'baz'])
+    {'foo': 1, 'baz': 'def'}
+
+    >>> select_keys({'foo': 1, 'bar': 2}, ['foo', 'baz'])
+    {'foo': 1}
+    """
+    return {
+        k: d[k]
+        for k in keys
+        if k in d
+    }
+
+
+def allowed_volume_name_subs():
+    """Helper to return the allowed volume name substitutions from the instance metadata"""
+    return select_keys(
+        boto.utils.get_instance_metadata(),
+        ['local-ipv4', 'public-ipv4']
+    )
 
 
 def retry(func):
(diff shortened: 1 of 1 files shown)


runtime/opt/taupage/init.d/10-prepare-disks.py
```diff
@@ -30,6 +31,29 @@ def zone():
...
+        k: d[k]
+        for k in keys
+        if k in d
+    }
+
+
+def allowed_volume_name_subs():
+    """Helper to return the allowed volume name substitutions from the instance metadata"""
+    return select_keys(
+        boto.utils.get_instance_metadata(),
```
aermakov-zalando: IMDS has hiccups sometimes. Does this retry?

aermakov-zalando: Although I guess it's used in other parts of the script anyway, so it won't be worse.

aermakov-zalando:
👍

runtime/opt/taupage/init.d/10-prepare-disks.py
a1exsh: I've also noticed that `find_volume` isn't annotated with `@retry`.  Should we add it to this PR or make a separate one just for that?

aermakov-zalando: Let's do it separately.

jrake-revelant:
👍

a1exsh:
Hm, I've just tried the actual `${local-ipv4}` substitution and it doesn't work because of the `-` in the middle.  Another PR will be needed to make it work.

PR #590, opened 2020-03-20
Fix IPv4 address substitution

a1exsh:
The hyphen inside key for substitution is not allowed.  Translate to keys with
underscore instead: `local_ipv4` and `public_ipv4`.

Signed-off-by: Oleksandr Shulgin <oleksandr.shulgin@zalando.de>


a1exsh:
Follow up for #589

a1exsh:
:+1:

Issue #592, opened 2020-04-01
docker repository has been deprecated


jeh0s:
`https://apt.dockerproject.org/repo` has been deprecated

```
Notice: Shutting down dockerproject.org APT and YUM repos 2020-03-31

Docker will be shutting down the deprecated APT and YUM repositories hosted at "dockerproject.org" and "dockerproject.com" on the 31st of March 2020.

We noticed that this project is referencing one of these repositories, and recommend updating to use the "download.docker.com" repository to prevent disruption.

More info: https://www.docker.com/blog/changes-dockerproject-org-apt-yum-repositories/ 
```

PR #591, opened 2020-03-24
Revert "Updated docker version (#586)"

aermakov-zalando:
Temporarily revert #586 due to unknown issues during the build.
Let's get the latest changes in and then try to update Docker again before the final deprecation.

PR #593, opened 2020-04-01
Change Docker repository (#592)

jeh0s:
`apt.dockerproject.org` has been deprecated, change new one.


mikkeloscar:
We had to revert our previous attempt https://github.com/zalando-stups/taupage/pull/591

We can pick this up when we have time to test with our internal pipelines.

